PR #28, opened 2014-11-04
Switch to jsonlite

yihui:
This might be a breaking change to the existing widget packages, but I think it should be easy for them to adapt to this change. My personal experience is that, among several other advantages, jsonlite covers the corner cases much better than RJSONIO. The default treatment of data frames also makes better sense to JS libraries (record/row-based instead of column-based). Another bonus point is @jeroenooms is extremely responsive, so we can bug him before he becomes a busy super star :-p

TODO:

- [x] https://github.com/bokeh/rbokeh/pull/56
- [x] https://github.com/jrowen/rhandsontable/issues/8


yihui:
A summary of this PR so far:

1. `toJSON()` was copied from the current dev version of shiny;
2. the attribute `TOJSON_ARGS` of `widget$x` is used to store possible arguments to be passed to `toJSON()`, which opens a door for widget authors to customize `toJSON()` arguments (such as `digits` #85);

The main issue that was holding up this PR was compatibility with shiny, which has been addressed by 1. But 2 may still bring incompatibility. I just thought about it again, and I think we have a few options:

1. Do nothing. Just document the potential risk if the widget is rendered in shiny. This makes sense to those who only use widgets in static documents. It is also fine if `TOJSON_ARGS` only contains cosmetic options such as `pretty = TRUE/FALSE`.
2. Make shiny recognize `TOJSON_ARGS`. Then we should not convert the `values` list as a whole object in shiny: https://github.com/rstudio/shiny/blob/babc59f85c/R/shiny.R#L514 We need to check the child elements of this list one by one, and convert them using `TOJSON_ARGS` individually. Then combine all child elements into the final JSON string.
3. Pre-generated JSON in [`htmlwidgets::shinyRenderWidget`](https://github.com/ramnathv/htmlwidgets/blob/a89c8696c/R/htmlwidgets.R#L282), and pass this JSON string to shiny, which will be double-encoded by `toJSON()` (e.g. `1:3` -> `[1, 2, 3]` -> `"[1, 2, 3]"`). Finally we just double-decode the JSON string (e.g. `JSON.parse(JSON.parse('"[1, 2, 3]"'))`) in https://github.com/ramnathv/htmlwidgets/blob/a89c8696c2/inst/www/htmlwidgets.js#L377 Alternatively, we can check in shiny if an element of the `values` has already been encoded (e.g. has the class `json`), and do not encode it one more time. Or we can ask @jeroenooms to add a method `setMethod("asJSON", "json", function(x, ...) as.character(x))`.

Personally I think 3 is a better way to go, since the serialization of the widget to JSON occurs only in htmlwidgets. Option 2 will handle the widget to `shiny:::toJSON()` instead, and we will have to be careful about maintaining the consistency of `toJSON()` between htmlwidgets and shiny. I'm wondering what you guys think.

Also Cc @wch.

R/utils.R
```diff
@@ -1,3 +1,26 @@
...
+  POSIXt = "ISO8601", UTC = TRUE, rownames = FALSE, keep_vec_names = TRUE
+) {
+  jsonlite::toJSON(
+    I(x), dataframe = dataframe, null = null, na = na, auto_unbox = auto_unbox,
+    digits = digits, use_signif = use_signif, force = force, POSIXt = POSIXt,
+    UTC = UTC, rownames = rownames, keep_vec_names = keep_vec_names, ...
+  )
+}
+
+if (requireNamespace('shiny')) local({
```
wch: Won't this code be run only at build time? I think you'd need to check at run time, or when the package is loaded.

yihui: This is a note to ourselves (authors of this htmlwidgets package) at build time. It may be confusing to end users if they see this warning, since they cannot really do anything about it.

wch:
I think the `TOJSON_ARGS` method is a bit fragile, and because it's tied to the unexported `htmlwidgets:::toJSON` function, it requires users to have knowledge of the inner workings of htmlwidgets and shiny.

I think it would be more robust and transparent to the user if they supplied their own toJSON function as an attribute like `json_serializer`. If they want the default behavior, they don't have do anything; if they want something different, they would be free to wrap up `jsonlite::toJSON` or even `RJSONIO::toJSON` however they want.

The other alternative, allowing users to serialize to JSON directly, and then pass that through untouched, also sounds reasonable.

yihui:
@wch The `json_serializer` approach was what I proposed in #30. It is a similar idea with the `TOJSON_ARGS` approach. The main issue is we need to deal with these special attributes in shiny.

jeroenooms:
The thing that worries me a bit is that there might be legitimate cases for double jsonning. For example if you have some multi layered system where the low-level framework (say, shiny) uses jsonlite for IO, but for a particular application of the framework, one of the values needs to be an actual JSON string. Then the application might use `jsonlite::toJSON` to generate the json string, and the framework wraps the string in a json blob using `jsonlite::toJSON` again. 

In this case the feature might be undesired and result in unexpected output. I'm not sure if it's a good idea to expect that the user will know it has to unclass the string for it to be treated as a string.

yihui:
@jeroenooms Note it has never been possible before to encode the output of `toJSON()`, because there was not `asJSON` method on the `json` class. The problem only lives in the future. I can think of two possible options:

1. Merge my current PR jeroenooms/jsonlite#88, and wait for the potential "legitimate cases for double encoding";
  - If users are happy with as.character(), we are fine;
  - If not, we can consider a new argument like `double_encode = TRUE/FALSE` (although I guess you are most likely not happy with it);
2. Give us a way to deal with the `json` class, e.g. by making the `force` argument a function instead of TRUE/FALSE, so we can decide what to do with this class by ourselves without you sponsoring the `json` class in jsonlite;

jcheng5:
@jeroenooms If you want to favor double jsonning, what about letting us opt-in using AsIs, so wrapping `I()` around a json object would let it be taken verbatim?

jeroenooms:
I guess we can do that, although `I()` is also used for the boxing so that could potentially conflict given that the `json` objects are by definition an atomic vector of length 1...

jeroenooms:
There is a more general case for including verbatim snippets in the json. For example if you want to include a JavaScript function in the object. In that case it is no longer json of course, but it might be useful.

jeroenooms:
Would it be an idea to use the `JS()` function for this?

jeroenooms:
Argh I keep replying to the wrong issue :)

yihui:
Okay, I think I have sorted out everything with this PR plus rstudio/shiny#795. Now htmlwidgets will pre-serialize the widget data to JSON for shiny, and shiny will not encode the JSON string again. This will require the latest development version of jsonlite: jeroenooms/jsonlite@6fd201b I have tested DT and it worked pretty well.

Customization of the JSON serializer is possible at two levels:

- customize the arguments of `jsonlite::toJSON()` via the `TOJSON_ARGS` attribute of the widget data (see https://github.com/rstudio/DT/blob/b9796a86/R/datatables.R#L185-L187 for an example);
- or customize the JSON serializer via the `TOJSON_FUNC` attribute;

It is up to the widget author whether he/she wants to expose such customization to users, e.g. one might want users to customize `digits` like this:

```r
foo_widget = function(..., digits = 7) {
  data = ...
  attr(data, 'TOJSON_ARGS') = list(digits = digits)
  createWidget(name, data, ...)
}
```

ramnathv:
This sounds great @yihui. What changes will widget authors have to make, in order to keep things working?

yihui:
I only tested DT and leaflet. I have not really found anything broken yet. The wrapper `shiny:::toJSON` is close enough to the behavior of `RJSONIO::toJSON`. After we merge this PR, we will have to notify all widget authors that we know and ask them to further test it.

jjallaire:
I just tested this with all of the widgets in the htmlwidgets showcase (http://www.htmlwidgets.org/) and they all worked save for metricsgraphics (which had a bunch of errors). @hrbrmstr could you try testing and let us know if we can do something or you need to do something to get things working. You can install the packages required to test with:

```r
devtools::install_github(c('jeroenooms/jsonlite', 'rstudio/shiny', 'yihui/htmlwidgets@jsonlite'))
```

jjallaire:
@timelyportfolio It would be great to know whether all of your widgets continue to work with the new JSON serializer.

timelyportfolio:
@jjallaire just ran through > 100 examples with widgets from building widgets and others.  all worked. great job everyone involved!

jjallaire:
Okay, let's wait to hear about what is required to resolve the issues in
metricsgraphics before we proceed.

hafen:
How soon are you guys planning on merging this?  I'd like to test with rbokeh.

yihui:
@hafen I guess we may merge it soon, but we will certainly wait until we hear no substantial problems from widget authors to make the CRAN release. Of course, we will have to coordinate with the shiny release as well.

@hrbrmstr @jjallaire I think I have figured out the problem. RJSONIO converts dates to numbers (number of days from 1970-01-01), and jsonlite uses the character representation instead. I can submit a PR to metricsgraphics to fix it.

jjallaire:
We are hoping to merge it soon. You should therefore ideally do your
testing ASAP!

timelyportfolio:
@hafen I checked with some `rbokeh` that I had and it worked great

jjallaire:
Okay, let me know once that PR is merged and I'll test again with the
showcase code.

hafen:
Ok - I just did some testing and everything works except for `ly_image`:

```
figure(xlim = c(0, 1), ylim = c(0, 1), title = "Volcano") %>%
  ly_image(volcano) %>%
  ly_contour(volcano)
```

I had to do some custom messing with the json there because RJSONIO didn't do what seemed natural.  Probably an easy fix but I'm on vacation so I'll get to it as soon as I can.

Go ahead with the merge when you're ready.

jjallaire:
cc'ing @bwlewis so he can be sure to test his widgets as well before we merge.

smartinsightsfromdata:
@yihui I just tried rhandsontable by @jrowen, implemented with htmlwidgets, and sadly nothing seem to work anymore (on rstudio as well as on shiny).

yihui:
@hafen Please see the one-liner fix at https://github.com/bokeh/rbokeh/pull/56

@smartinsightsfromdata I'll take a look at rhandsontable soon.

jrowen:
For `rhandsontable`, I think the `JSON.parse` calls in rhandsontable.js are causing the problem.  If I remove these, the widget renders.  In the R code `jsonlite::toJSON` is being used to pass some data and config parameters to the widget.  With this update, maybe there's now a better way to handle this.

bwlewis:
This branch seems is working fine with threejs so no problem from me if you merge this.

The threejs widget uses the rjson package on its own in a few places. I will work on moving that over to jsonlite this weekend, and expect to push threejs to cran maybe Sunday night.

jeroenooms:
@yihui I think @jrowen's problem is what [I was afraid of](https://github.com/jeroenooms/jsonlite/pull/88) with verbatim json includes. When the user legitimately wants to include a json string to parse it manually on the client, you will get unexpected results when the json gets included verbatim.

@jrowen Try using `unclass(jsonlite::toJSON(x))` in your code.

bwlewis:
OK, I spoke too soon. Something else is wrong with this PR, perhaps unrelated to the use of jsonlite. My threejs shiny examples are broken, with undefined arguments in the JavaScript render function. Those examples work fine with the released version of htmlwidgets.

I'll try to track down exactly what is causing this problem, but please hold off merging this.

jrowen:
@jeroenooms The `unclass` suggestion fixed the error.  Thanks.  

I'm running now into an "Input to asJSON(keep_vec_names=TRUE)..." message when passing a `data.frame` to the widget using `unclass(jsonlite::toJSON(data, na = "string"))` (no message with `matrix`).

yihui:
@jeroenooms @jrowen Wait a minute. I guess we are heading to the wrong direction by using `unclass()` -- we should have just removed `JSON.parse()` instead. Let me experiment with it and get back to you. Jeroen, I believe this is a perfect case why `json_verbatim = TRUE` is a better default, contrary to what you were thinking, because there is really no need to double encode the data object, _and_ double decode it. Encoding and decoding should be done only once.

timelyportfolio:
@bwlewis might your issues be caused by `auto_boxing`?

timelyportfolio:
@ThomasSiegmund , you might be interested in this discussion.  Could you try your D3TableFilter with this pull?

```
devtools::install_github(c('jeroenooms/jsonlite', 'rstudio/shiny', 'yihui/htmlwidgets@jsonlite'))
```

yihui:
@bwlewis I sent a PR to you and tested your shiny examples under https://github.com/bwlewis/rthreejs/tree/master/inst/examples and other static examples in the documentation. I did not see any issues. Please let me know you still have any problems (a reproducible example will be helpful).

@jrowen I verified what I was thinking: your use of `toJSON()` + `JSON.parse()` was mostly due to the fact that RJSONIO did not work well enough, so you have to hack around using jsonlite. Now you don't need the dark voodoo any more, and can enjoy a clean interface between R and JavaScript without taking care of encoding/decoding the data manually.

bwlewis:
@timelyportfolio @yihui 
Thanks Yihui, your threejs PR is merged and works great. No issues anymore for me!

yihui:
@bwlewis Great!! I think @timelyportfolio was correct that `auto_unbox` was the reason. We changed the defaults of `jsonlite::toJSON()` by using a wrapper function `shiny:::toJSON()`, and `auto_unbox = FALSE` was changed to `TRUE`.

@timelyportfolio I tested @ThomasSiegmund's D3TableFilter package, and the examples still work fine.

ThomasSiegmund:
Hi all,

I can confirm that D3TableFilter works fine. I get a warning about a named
vector not beeing supported in future versions of jsonlite. Will change my
code accordingly.

Thanks for taking care

Thomas

smartinsightsfromdata:
I've re-tested my widgets:
- rpivotTable
- rdatamaps
- rd3pie

And they all work fine, beside the warning below in rdatamaps (already corrected by @yihui in rhandsontable, so I have an idea where to look to fix this):
```R
Input to asJSON(keep_vec_names=TRUE) is a named vector. In a future version of jsonlite, this option will not be supported, and named vectors will be translated into arrays instead of objects. If you want JSON object output, please use a named list instead. See ?toJSON
```
Many thanks to everyone: it is really rewarding been part of this community.

yihui:
@ThomasSiegmund @smartinsightsfromdata You just need to find out where the named vector is created, and turn it into a named list instead. Here is the fix I provided to rhandsontable: https://github.com/yihui/rhandsontable/commit/67462eb9156edc480572f2f89eeed7d8587595f6

jeroenooms:
FYI the new jsonlite is on CRAN already.

DESCRIPTION
```diff
@@ -13,7 +13,7 @@ License: MIT + file LICENSE
 VignetteBuilder: knitr
 Imports:
     htmltools (>= 0.2.6),
-    RJSONIO,
+    jsonlite,
```
jjallaire: Is there a version requirement for jsonlite or will any version work?

NAMESPACE
```diff
@@ -10,5 +10,4 @@ export(scaffoldWidget)
 export(shinyRenderWidget)
 export(shinyWidgetOutput)
 export(sizingPolicy)
-import(RJSONIO)
 import(htmltools)
```
jjallaire: do we need import(jsonlite) here?

DESCRIPTION
ramnathv: I think we should specify that it is the latest version. (>= 0.9.16.9000)

NAMESPACE
ramnathv: We should not need this in the NAMESPACE, since all `jsonlite` functions are being used as `jsonlite::toJSON` etc.

DESCRIPTION
ramnathv: Going by the CRAN version, this should be `jsonlite (>= 0.9.16)`

yihui: Good point. Done.

NAMESPACE
yihui: Ramnath is correct.

PR #61, opened 2014-12-31
Added coffeewheel to the sample packages list

armish:
`coffeewheel` is yet another `htmlwidgets`-based implementation. Thought it would be good to have it in the example packages list.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ There are already several R packages based on **htmlwidgets**, including:
 * [sparkline](https://github.com/htmlwidgets/sparkline) --- Small inline charts
 * [DT](http://rstudio.github.io/DT/) --- Tabular data via DataTables
 * [rthreejs](https://github.com/bwlewis/rthreejs) -- Interactive 3D graphics
+* [cofeewheel](https://github.com/bwlewis/rthreejs) -- Interactive hierarchical data visualization with D3 Sunburst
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,6 +14,7 @@ There are already several R packages based on **htmlwidgets**, including:
 * [sparkline](https://github.com/htmlwidgets/sparkline) --- Small inline charts
 * [DT](http://rstudio.github.io/DT/) --- Tabular data via DataTables
 * [rthreejs](https://github.com/bwlewis/rthreejs) -- Interactive 3D graphics
+* [cofeewheel](https://github.com/bwlewis/rthreejs) -- Interactive hierarchical data visualization with D3 Sunburst
```
jjallaire: The link here is to rthreejs rather than coffeewheel

jjallaire:
From looking at the source code of coffeewheel it appears as if there may still be work to do in sizing. You default to 400px however depending on the container this may or may not actually work well. If you  need to force a square size then you should do that at the last possible instant (e.g. if the framework calls you with 600x500 you could just render at 500x500).

armish:
Hi JJ,

Thanks for reviewing the code. Yes, you are right: I can improve handling the resizing and as you suggested, I will fix the ambuigity with the size (height, width and forcing it to be a square). Let me make the changes in the code first and I will make another pull request with these new changes. Feel free to ignore this one for now.

Best,
-- Arman

README.md
armish: argh! Apparently I forgot to change the URL after I copied and pasted. My bad, will be fixed in the next request.

Thanks,
-- Arman

PR #68, opened 2015-01-09
add as.iframe function to allow embedding widget html as iframes.

ramnathv:
This PR allows html widgets to be embedded as iframes. It works by saving the widget to html using `saveWidget` and then using `htmltools` to embed it as an iframe. 

By default, it saves the widget html to file using `selfcontained = TRUE` and then uses the [srcdoc](http://www.w3schools.com/TAgs/att_iframe_srcdoc.asp) attribute to embed it as an inline iframe. This allows the resulting html document to be selfcontained. The `srcdoc` attribute is not supported by IE, and for some reason does not appear to be supported in the Rmd previewer as well.

@jjallaire @jcheng5 I tested this with a few widgets and it seems to be working well. If you could go over the PR and get back with comments/feedback, I can roll this to completion.


R/utils.R
```diff
@@ -122,3 +122,39 @@ shouldEval <- function(options) {
...
+#' @inheritParams saveWidget
+#' @param width width of iframe
+#' @param height height of iframe
+#' @param border-width width of the borders. default is 0px
+#' @param ... css styles to add to the iframe
+#' @export
+as.iframe <- function(widget, file = NULL, selfcontained = TRUE,
+    libdir = NULL, width = '100%', height = 500, `border-width` = '0px', ...){
+  if (is.null(file)){
+    file <- basename(tempfile(fileext = '.html'))
```
jcheng5: 1. Why `basename`?
2. Does selfcontained=FALSE make sense? I guess yes iff the libdir is passed in?
3. If the file is null, do we also need `on.exit(unlink(file), add = TRUE)`?

R/utils.R
```diff
@@ -122,3 +122,39 @@ shouldEval <- function(options) {
...
+  }
+  htmlwidgets::saveWidget(widget,
+    file = file, selfcontained = selfcontained, libdir = libdir
+  )
+  styles = list(`border-width` = `border-width`, ...)
+  style = paste(paste(names(styles), styles, sep = ':'), collapse = ';')
```
jcheng5: I'd change to `paste(paste(names(styles), styles, sep = ':'), ';')` so that the last attribute is semicolon-terminated as well; and perhaps `...` should be interpreted not as style rules, but as tag attributes. If you make both these changes, then you could still add additional styles by adding another `style` attribute, but still be able to do other attributes:

`as.iframe(widget, style="border-color: green;", id="widget_iframe1")`

This works because duplicate attributes are concatenated together:

```r
> div(style="font-family:Times;", style="color:blue;")
<div style="font-family:Times; color:blue;"></div> 
```

R/utils.R
```diff
@@ -122,3 +122,39 @@ shouldEval <- function(options) {
...
+      srcdoc = paste(readLines(file), collapse = '\n'),
+      width = width, height = height, style = style
+    )
+  } else {
+    list(
+      src = file,
+      width = width, height = height, style = style
+    )
+  }
+  do.call(htmltools::tags$iframe, content)
```
jcheng5: I think [according to Hadley's advice](http://rpubs.com/hadley/do-call2) this should be `do.call(htmltools::tags$iframe, list(quote(content)))`, bizarre as that seems.

R/utils.R
```diff
@@ -122,3 +122,39 @@ shouldEval <- function(options) {
...
+    file <- basename(tempfile(fileext = '.html'))
+  }
+  htmlwidgets::saveWidget(widget,
+    file = file, selfcontained = selfcontained, libdir = libdir
+  )
+  styles = list(`border-width` = `border-width`, ...)
+  style = paste(paste(names(styles), styles, sep = ':'), collapse = ';')
+  content <- if (selfcontained){
+    list(
+      srcdoc = paste(readLines(file), collapse = '\n'),
```
jcheng5: Wow, that's wild.

R/utils.R
ramnathv: 1. The filename matters only for the option `selfcontained = FALSE` in which case it makes more sense to save it relative to the directory of the `rmd` file. 
2. Even without `libdir`, I think `selfcontained = F` makes sense, since it will save the widget html to  file and link dependencies from the default value of `libdir`, which I think is the same as the widget name.
3. I don't think we want to unlink the file, especially when `selfcontained = FALSE`, since it is referred to by the `iframe`. We can unlink it for `selfcontained = TRUE`.

R/utils.R
ramnathv: You are right that `...` refers to tag attributes. I will update the documentation to reflect that. The change you suggested makes sense and I will incorporate that as well.

R/utils.R
ramnathv: I would never have figured this one out! I will trust @hadley on this :smile:

R/utils.R
ramnathv: I learnt the `srcdoc` trick while working on `rCharts`. For some reason, iframes using the `srcdoc` attribute don't seem to display in the rstudio preview window. Thoughts?

ramnathv:
I messed this one up as I made the updates. So I am closing this PR. I will start a new one.

PR #69, opened 2015-01-10
add as.iframe function to embed widgets as iframes in rmd documents

ramnathv:
This PR allows html widgets to be embedded as iframes. It works by saving the widget to html using saveWidget and then using htmltools to embed it as an iframe.

By default, it saves the widget html to file using selfcontained = TRUE and then uses the srcdoc attribute to embed it as an inline iframe. This allows the resulting html document to be selfcontained. The srcdoc attribute is not supported by IE, and for some reason does not appear to be supported in the Rmd previewer as well.

According to suggestions from @jcheng5, I have made the following changes:

1. added semicolon terminator to the last tag
2. reworded the documentation for the ... parameter to reflect that they are tag attributes.
3. use the `list(quote(x))` hack in `do.call`.


ramnathv:
@jcheng5 The `list(quote(content))` hack in the `do.call` statement seems to backfire here. So I have reverted it back to just `content`. If you can take a look, that would be great.

jcheng5:
Yeah sorry I misunderstood the rpubs document. I coincidentally have encountered the same issue in my own code today and I think I understand it better, will let you know when I figure it out.

R/utils.R
```diff
@@ -122,3 +122,43 @@ shouldEval <- function(options) {
...
+      srcdoc = paste(readLines(file), collapse = '\n'),
+      width = width, height = height, style = style
+    )
+  } else {
+    list(
+      src = file,
+      width = width, height = height, style = style
+    )
+  }
+  do.call(htmltools::tags$iframe, content)
```
jcheng5: Actually, you could avoid any question by just doing `htmltools::tag("iframe", content)`. That's all `tags$iframe` is doing anyway.

ramnathv: Done.

timelyportfolio:
Used this for my post for week 2 and in a couple other applications, and it worked great.

timelyportfolio:
is there any reason not to merge this?  I have used on all my posts for buildingwidgets with no problems.

ramnathv:
@jcheng5 can you take a look at this PR and merge when you feel it is good to go?

timelyportfolio:
Just for reference I have used at least 200 times with all sorts of widgets (good and bad) with no problems.

jcheng5:
Do you mind just adding a note that it doesn't work with the RStudio Viewer pane nor Rmd previewer?

Other than that, LGTM.

ramnathv:
Ah, I didn't realize that it doesnt work in the Viewer/Previewer. @jcheng5 is that a limitation of the browser in use in the viewer/previewer, or something about iframes? Ideally, I would like to fix this before we merge this change in.

jcheng5:
cc @jjallaire @jmcphers

It looks to me like the `<iframe srcdoc>` attribute isn't supported in the browser engine RStudio is using, at least on Linux (I'm 90% sure the same would be true for Windows as it also uses Qt, and not sure what happens on the Mac as it basically uses Safari). I say this because passing selfcontained=FALSE to the as.iframe call makes it work.

Repro:

```r
devtools::install_github("ramnathv/htmlwidgets@feature/as-iframe-v2")
devtools::install_github("rstudio/leaflet")
library(leaflet)
library(htmlwidgets)
as.iframe(leaflet() %>% addTiles())
```

On second look, selfcontained=FALSE works in Rmd but not from the console (e.g. `htmltools::browsable(as.iframe(leaflet() %>% addTiles(), selfcontained=FALSE))`) as the iframe's src file isn't relocated. And selfcontained=FALSE doesn't work in Shiny either.

timelyportfolio:
This is sort of typical use for me https://github.com/timelyportfolio/buildingwidgets/blob/gh-pages/week04/index.Rmd, since the blog does not provided `iframe src` either and it previews nicely in RStudio Viewer.

ramnathv:
Ok. I am able to repro the issue. I think there are two different issues here.

First, the issue of `<iframe srcdoc=...></iframe>`. It is well known that the `srcdoc` attribute does not play well with some browsers. So all problems on that front are not something we can resolve.

Coming to the use of `<iframe src-...></iframe>`, the widget html is saved to file and then linked. It works great in static Rmd since the file is saved relative to the directory in which the Rmd file resides. While using in the console, it does not work, since the linked html needs to be copied to a path relative to where the parent html file is being copied for rendering with rstudio viewer, and I am not sure how that path can be accessed. 

It would be nice to make `as.iframe` play well under multiple settings including Shiny, although my primary intended use was with static Rmds. Let us keep this PR open till we are able to resolve things or agree on the limitations and document them.

timelyportfolio:
any way we can just accept the limitations and move forward with this even if it is not exported?  In my view, there is no way around the limitations and little progress to resolve.

ramnathv:
I agree @timelyportfolio. I will add some documentation to `as.iframe` to make it really clear what these limitations are, so that it is transparent to users.

ramnathv:
Superceded by #154

Issue #71, opened 2015-01-12
Correct idiom to inject static JSON Data


schnee:
I'm attempting to "simply" draw a map using D3 and D3 Projections via htmlwidgets. My question is how to "inject" the shapes data, in the "htmlwidgets" way.

Currently, my renderValue looks like:

```javascript
  renderValue: function(el, x, instance) {

    var path = d3.geo.path()
    .projection(instance.projection);

    // select the svg element and remove existing childern
    var svg = d3.select(el).select("svg");
    svg.selectAll("*").remove();
        
    d3.json("htmlwidgets/lib/shapes/world-110m.json", function(error, world) {
      svg.insert("path", ".trip")
          .datum(topojson.feature(world, world.objects.land))
          .attr("class", "land")
          .attr("d", path);
      
      svg.insert("path", ".trip")
          .datum(topojson.mesh(world, world.objects.countries, function(a, b) { return a !== b; }))
          .attr("class", "boundary")
          .attr("d", path);
  });
},
```
My question centers on this line:

 d3.json("htmlwidgets/lib/shapes/world-110m.json", function(error, world) {

What is the idiomatic way to, in general, provide static data like this? 

My yaml looks like:
```
dependencies:
- name: d3-geo-projection
  version: 0.2.12
  src: htmlwidgets/lib/d3-geo-projection
  script: d3-geo-projection.js
- name: d3
  version: 3.5.3
  src: htmlwidgets/lib/d3
  script: d3.min.js
- name: topojson
  version: 1.6.18
  src: htmlwidgets/lib/topojson
  script: topojson.js
```

jjallaire:
@ramnathv Would the attachment work that we did cover this use case?

ramnathv:
Yes. The attachment stuff would work. Let me cook up a simple example that can go into the docs.

jjallaire:
Okay, you can add it as a section of the "Advanced" article

ramnathv:
@schnee I will be posting a more detailed article on how to handle extra dependencies within the `htmlwidgets` framework. But here is a quick version of it.

__Step 1__

Include `shapes` as a dependency in the YAML

```yaml
dependencies:
  - name: 'shapes'
    version: 0.1.0
    src: 'htmlwidgets/lib/shapes'
    attachment: {"world": 'world-110m.json'}
```

__Step 2__

Access it in your javascript binding

```js
// first argument refers to dependency name and the second argument refers to the name of the 
// attachment you are trying to access.
[...]

PR #83, opened 2015-02-18
Easier specification of external file dependencies

ramnathv:
This PR makes it easy for package developers and users to specify file dependencies for their widgets. Here is an example:

```r
# install_github('htmlwidgets/datamaps')
library(datamaps)
datamaps(
  scope = 'pcs',
  geographyConfig = list(dataUrl = htmlwidgets:::attachment("data/pcs.json"))
)
```

It uses a similar mechanism to the one used by the `JS` function to mark objects as javascript literals. The resolution of attachment urls is carried out using the `HTMLWidgets.getAttachmentUrl` function. 

@jcheng5 can you take a look at this whenever you have time. @jjallaire @timelyportfolio @yihui if you can test this functionality with any of your widgets and provide me with feedback/comments, that would be very useful.

### DIFF
--- a/R/fileDependency.R
+++ b/R/fileDependency.R
@@ -1,42 +1,26 @@
-#' Define a file dependency that can be accessed by the client using javascript.
+#' Create a file dependency that can be accessed by the client using javascript.
 #' @export
 fileDependency <- function(filename, version = '0.0.1'){
   htmltools::htmlDependency(
     name = basename(tools:::file_path_sans_ext(filename)),
     version = version,
-    src = normalizePath(dirname(filename)),
+    src = dirname(filename),
     attachment = basename(filename)
   )
 }
 
+#' Mark a string as an attachment
+#' @export
 attachment <- function(x){
-  structure(x, class = unique(c("ATTACHMENT", oldClass(x))))
+  structure(normalizePath(x), class = unique(c("ATTACHMENT", oldClass(x))))
 }
 
-attachmentEvals <- function(list) {
-  evals <- which(unlist(shouldEval2(list)))
-  I(evals)  # need I() to prevent RJSONIO::toJSON() from converting it to scalar
+attachmentDeps <- function(list) {
+  attachments = rapply(list, function(y){y}, classes = 'ATTACHMENT')
+  deps = lapply(attachments, fileDependency)
+  attachments = lapply(as.list(attachments), function(x){
+    basename(tools::file_path_sans_ext(x))
+  })
+  list(attachments = attachments, deps = deps)
 }
 
-#' JSON elements that are character with the class JS_EVAL will be evaluated
-#'
-#' @noRd
-#' @keywords internal
-shouldEval2 <- function(options) {
-  if (is.list(options)) {
-    if ((n <- length(options)) == 0) return(FALSE)
-    # use numeric indices as names (remember JS indexes from 0, hence -1 here)
-    if (is.null(names(options)))
-      names(options) <- seq_len(n) - 1L
-    # Escape '\' and '.' by prefixing them with '\'. This allows us to tell the
-    # difference between periods as separators and periods that are part of the
-    # name itself.
-    names(options) <- gsub("([\\.])", "\\\\\\1", names(options))
-    nms <- names(options)
-    if (length(nms) != n || any(nms == ''))
-      stop("'options' must be a fully named list, or have no names (NULL)")
-    lapply(options, shouldEval2)
-  } else {
-    is.character(options) && inherits(options, 'ATTACHMENT')
-  }
-}
--- a/R/fileDependency.R
+++ b/R/fileDependency.R
@@ -1,0 +1,42 @@
+#' Define a file dependency that can be accessed by the client using javascript.
+#' @export
+fileDependency <- function(filename, version = '0.0.1'){
+  htmltools::htmlDependency(
+    name = basename(tools:::file_path_sans_ext(filename)),
+    version = version,
+    src = normalizePath(dirname(filename)),
+    attachment = basename(filename)
+  )
+}
+
+attachment <- function(x){
+  structure(x, class = unique(c("ATTACHMENT", oldClass(x))))
+}
+
+attachmentEvals <- function(list) {
+  evals <- which(unlist(shouldEval2(list)))
+  I(evals)  # need I() to prevent RJSONIO::toJSON() from converting it to scalar
+}
+
+#' JSON elements that are character with the class JS_EVAL will be evaluated
+#'
+#' @noRd
+#' @keywords internal
+shouldEval2 <- function(options) {
+  if (is.list(options)) {
+    if ((n <- length(options)) == 0) return(FALSE)
+    # use numeric indices as names (remember JS indexes from 0, hence -1 here)
+    if (is.null(names(options)))
+      names(options) <- seq_len(n) - 1L
+    # Escape '\' and '.' by prefixing them with '\'. This allows us to tell the
+    # difference between periods as separators and periods that are part of the
+    # name itself.
+    names(options) <- gsub("([\\.])", "\\\\\\1", names(options))
+    nms <- names(options)
+    if (length(nms) != n || any(nms == ''))
+      stop("'options' must be a fully named list, or have no names (NULL)")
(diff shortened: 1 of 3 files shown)


R/fileDependency.R
```diff
@@ -0,0 +1,26 @@
...
+    version = version,
+    src = dirname(filename),
+    attachment = basename(filename)
+  )
+}
+
+#' Mark a string as an attachment
+#' @export
+attachment <- function(x){
+  structure(normalizePath(x), class = unique(c("ATTACHMENT", oldClass(x))))
```
jcheng5: Why `oldClass` vs. `class` here?

ramnathv: I just used the same code as used in `htmlwidgets:::JS`. I should admit I did not understand why `oldClass` was used there in the first place :smile:

ramnathv: I should also probably add a `file.exists(x)` check in the `attachment` function to ensure that incorrect paths are caught early on.

R/fileDependency.R
```diff
@@ -0,0 +1,26 @@
+#' Create a file dependency that can be accessed by the client using javascript.
+#' @export
+fileDependency <- function(filename, version = '0.0.1'){
+  htmltools::htmlDependency(
+    name = basename(tools:::file_path_sans_ext(filename)),
+    version = version,
+    src = dirname(filename),
+    attachment = basename(filename)
+  )
```
jcheng5: I'm not sure `htmlDependency` is the right substrate to build this on. The reason for this is that htmltools and Shiny assume that once a dependency is satisfied, it's satisfied for good. So in your example, if in a Shiny app, pcs.json changes between one run and the next (and in fact if that's the reason it's changing) then the changes would be ignored.

You could fix this by using a random value for the htmlDependency's name, I guess? That would have an unfortunate side effect of never letting go of old values (or at least references to them)... though I think that's still better than not picking up new values.

jcheng5: I wonder... would there be a way to have htmlwidgets instead of resolving the dependency, just insert a link directly into the page?

ramnathv: Good points @jcheng5 . I did not think from this angle. I assumed the external file dependencies to be static resources, and this PR was an attempt to package the boilerplate code into something more automatic.

What you suggest here, would make this mechanism even more powerful, by allowing dynamic external dependencies. Your first comment seems like a simple enough fix that would give us reasonable mileage. However, if there is a way to insert the link directly in the page, I think it would work even better.

R/fileDependency.R
```diff
@@ -0,0 +1,29 @@
...
+}
+
+#' Mark a string as an attachment
+#' @export
+attachment <- function(x){
+  if (!file.exists(x)){
+    stop("The attachment ", x, " does not exist")
```
yihui: How about just leaving it to `normalizePath(mustWork = TRUE)`?

R/fileDependency.R
```diff
@@ -0,0 +1,29 @@
...
+  }
+  structure(normalizePath(x), class = unique(c("ATTACHMENT", oldClass(x))))
+}
+
+attachmentDeps <- function(list) {
+  attachments = rapply(list, function(y){y}, classes = 'ATTACHMENT')
```
yihui: `identity` == `function(y){y}`

R/fileDependency.R
ramnathv: Good suggestion! I will make the change.

R/fileDependency.R
ramnathv: I keep forgetting that this function exists. Thanks for the reminder!

nsh87:
This would be a most-welcome feature. I have been looking at [this issue](https://github.com/ramnathv/htmlwidgets/issues/71) in my attempt to attach a file dependency that gets created in one of my functions before `htmlwidgets::createWidget`. Thus far I have not been successful in  making the file accessible through the browser.

jcheng5:
@nsh87 Does this data need to be packaged as a separate file? Could it perhaps be included as part of the createWidget's `x` value?

nsh87:
@jcheng5 It's an XML file, so I don't think it can be included with `x` since that goes through `toJSON`. Outside of `htmlwidgets`, the JS library I'm using reads the data from an XML file, typically through AJAX. What I'm attempting to do is to get the file into `<head>` as an attachment and try to read it from there, but no luck attaching it so far.

```R
radial_phylo <- function(user-data, width=NULL, height=NULL) {

   # Do some stuff to save an XML file to xmlFilePath

   phyloxml <- htmltools::htmlDependency(
     name = 'phyloxmls',
     version = '1.0',
     src = c(file=xmlFilePath)
     attachment = list(xml = xmlFilePath)  # Pretty sure this isn't right
   )

  # create widget
  htmlwidgets::createWidget(
    name = 'radial_phylo',
    x,
    width = width,
    height = height,
    htmlwidgets::sizingPolicy(
      viewer.padding = 0,
      viewer.suppress = suppressViewer,
      browser.fill = TRUE
    ),
    package = 'receptormarker',
    dependencies = phyloxml
  )
}
```

jcheng5:
I think that's pretty close--in the JavaScript side you should then be able to call `HTMLWidgets.getAttachmentUrl("phyloxmls", "xml")` and get back a URL...?

nsh87:
That JS call works, but I'm not sure if the reference is right. If I inspect the attachment in the HTML and click on the link it's just a blank XML file (and `xmlFilePath` is not blank).

```JavaScript
var phyloxmlfile = HTMLWidgets.getAttachmentUrl('phyloxml', 'xml');
// phyloxmlfile = "lib/phyloxml-1.0//Users/Nikhil/dev/receptormarker/receptormarker/phyloxml-69647efd9cfd.xml"
```

nsh87:
Plus I don't have a `"inst/htmlwidgets/lib/phyloxml-1.0"` folder. Judging by the fact that it's looking in `"lib/phyloxml-1.0"`, it looks like it's expecting the XML asset to exist in my package, but the XML file is user-generated.

jcheng5:
This chunk:

```r
   phyloxml <- htmltools::htmlDependency(
     name = 'phyloxmls',
     version = '1.0',
     src = c(file=xmlFilePath)
     attachment = list(xml = xmlFilePath)  # Pretty sure this isn't right
   )
```

Should be this:

```r
   phyloxml <- htmltools::htmlDependency(
     name = 'phyloxmls',
     version = '1.0',
     src = c(file=dirname(xmlFilePath))
     attachment = list(xml = basename(xmlFilePath))  # Pretty sure this isn't right
   )
```

Note that the entire contents of `dirname(xmlFilePath)` will be copied along with `xmlFilePath`, so if that includes a lot of extraneous stuff, copy your XML file to an empty temp directory and then pass that temp directory as your `src` (and still the `basename()` as the attachment value).

nsh87:
Awesome, that did the trick. Lines 8 and 9 from the top of this pull request should've been good hints for me. Thanks.

jcheng5:
OK, great, glad to hear it. :)

ramnathv:
@jcheng5 Should we revisit this PR (i need some cleanup so it is mergeable)? All this PR does is provide a wrapper around the dependency API that you have put in. So maybe one way to go about this is to merge this in, acknowledging the limitations you have pointed out, and then once we have a more complete way of doing this, update the mechanism. Let me know what you think.

timelyportfolio:
:+1: +1

jcheng5:
I think it might be a simple tweak to get it working for dynamic values. Give me a minute...

ramnathv:
@jcheng5 I look forward to your magic :+1:

jcheng5:
OK, I pushed some changes. Some notes:

1. I took dependencies on digest and bitops, hope that's OK.
2. Should `fileDependency` be exported? I'm not sure how one would use it correctly outside of `attachment()` (although perhaps that's only the case after my change).
3. The underlying mechanism for traversing the data object based on key, will only work if neither the attachments nor their parent lists are ever unnamed lists (nor may their names have periods in them).
4. I didn't write docs/examples, that would need to be done too before merging.

ramnathv:
Thanks @jcheng5. I will play with your PR and do some testing. I am okay with the dependencies on `digest` and `bitops`. I will be happy to write up docs/examples before we merge.

PR #85, opened 2015-03-02
Add digits argument to createWidget

kbroman:
- In cases where there is a lot of numeric data, it'd be nice to pass a
   digits argument to `toJSON()`, to reduce the number of digits printed.

 - Simplest thing seemed to be to make it an argument to `createWidget,`
   add it to resulting structure, and then use it within `toHTML()`.


ramnathv:
This PR is related to [this](https://github.com/ramnathv/htmlwidgets/pull/43). Providing more control over number of digits is important, but the key question is what should be the entry point to control it.

jcheng5:
This is a good starting point. As written, it won't work with Shiny though.
I need to think a bit about how we might control that...

PR #93, opened 2015-03-30
Export staticRender

hafen:
@jcheng5, @jjallaire - thanks for helping with my issue last week.  With minimal effort this morning I was able to embed htmlwidgets in my application.  However, it required this one small change - exporting `staticRender()` in htmlwidgets.js.  Hopefully this is an acceptable change.

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -422,14 +422,14 @@
     if (!scheduleStaticRenderTimerId) {
       scheduleStaticRenderTimerId = setTimeout(function() {
         scheduleStaticRenderTimerId = null;
-        staticRender();
+        window.HTMLWidgets.staticRender();
       }, 1);
     }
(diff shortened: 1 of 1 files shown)


jjallaire:
I'm fine with this. @jcheng5 or @ramnathv Is there any reason we'd want to expose this API in a different fashion?

ramnathv:
@jjallaire I don't see any pitfalls of exposing this function, other than a situation, where we want to change the signature of the function to support things like user-defined post-render or pre-render hooks. For now, I think we can expose it, but leave it undocumented so that we have the flexibility to do things later on.

hafen:
Thanks!

jcheng5:
Sorry for the late reply, it's a good change. This function can't have arguments anyway since it's primarily designed to be invoked in an onDOMContentReady event handler.

Issue #65, opened 2015-01-05
how to handle callback / post-render behaviors ?


timelyportfolio:
Do we expect `htmlwidgets` to handle `callback` or other post-render behaviors, or should we expect widget authors to handle on their end?  `dygraphs` handles very nicely.  I wonder though given potential lack of Javascript knowledge on the part of some widget authors if we can establish some best practices for this.  See https://github.com/rich-iannone/DiagrammeR/pull/10.

jjallaire:
I think widget authors need to handle this. We've given them the JS
function and some examples of it's use, hopefully that (and pull requests
from us!) will be enough to get people pointed in the right direction.

timelyportfolio:
I agree for now, but the one situation where dealing with this at the individual widget level might be problematic will be a page with multiple `htmlwidgets` from multiple packages where the package authors differ in their implementation or do not implement.  Then, finding when all widgets are built will be troublesome.

PR #94, opened 2015-03-30
Feature/pre render hook

ramnathv:
This PR adds a `preRenderHook` argument to `createWidget` to support manipulations of the widget object, just prior to rendering. It was motivated by [this thread](https://github.com/bokeh/rbokeh/issues/25#issuecomment-87831557)

Here is a simple example that explains how it works

```R
devtools::create('newpkg')
setwd('newpkg')
htmlwidgets::scaffoldWidget('testwidget')
```

In the createWidget line of `testwidget`, add the following line

```
createWidget(
  x,
  ...,
  preRenderHook = function(x){ 
    x$x$message = paste("Hi", x$x$message) 
  })
)
```

If you build the package now and run  `testwidget("R"), it will render `Hi R`. It works with Shiny as well.

In the process of adding this feature, I also refactored `widget_data` and `shinyRenderWidget` by spinning common code off to a function called `createPayload`. This would make it easier for changes made to payload processing to permeate to both the static and shiny cases.

@jcheng5, @jjallaire, @yihui, @timelyportfolio Let me know if you have a chance to test this out, and any feedback/comments.

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -158,8 +158,9 @@ widget_data <- function(x, id, ...){
 #'@param elementId Use an explicit element ID for the widget (rather than an
 #'  automatically generated one). Useful if you have other JavaScript that needs
 #'  to explicitly discover and interact with a specific widget instance.
-#'@param preRenderHook A function to be run on the widget payload x, prior to
-#'  rendering.
+#'@param preRenderHook A function to be run on the widget, just prior to
+#'  rendering. It accepts the entire widget object as input, and should return
+#'  a modified widget object.
 #'
 #'@details
 #'
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -122,12 +122,9 @@ widget_dependencies <- function(name, package){
 # Generates a <script type="application/json"> tag with the JSON-encoded data,
 # to be picked up by htmlwidgets.js for static rendering.
 widget_data <- function(x, id, ...){
-  if (!is.null(x$preRenderHook)){
-    x$x = x$preRenderHook(x$x)
-  }
-  evals <- JSEvals(x$x)
+  payload <- createPayload(x)
   tags$script(type="application/json", `data-for` = id,
-    HTML(toJSON(list(x = x$x, evals = evals), collapse = "", digits = 16))
+    HTML(toJSON(payload, collapse = "", digits = 16))
   )
 }
 
@@ -276,20 +273,25 @@ shinyRenderWidget <- function(expr, outputFunction, env, quoted) {
         instance$elementId, "\"; Shiny doesn't use them"
       )
     }
-    x <- .subset2(instance, "x")
-    if (!is.null(instance$preRenderHook)){
-      x <- instance$preRenderHook(x)
-    }
     deps <- .subset2(instance, "dependencies")
     deps <- lapply(
       htmltools::resolveDependencies(deps),
       shiny::createWebDependency
     )
-    evals = JSEvals(x)
-    list(x = x, evals = evals, deps = deps)
+    payload = modifyList(createPayload(instance), list(deps = deps))
   }
 
   # mark it with the output function so we can use it in Rmd files
   shiny::markRenderFunction(outputFunction, renderFunc)
+}
+
+# Helper function to create payload
+createPayload <- function(instance){
+  x <- .subset2(instance, "x")
+  if (!is.null(instance$preRenderHook)){
+    x <- instance$preRenderHook(x)
+  }
+  evals = JSEvals(x)
+  list(x = x, evals = evals)
 }
 
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -122,6 +122,9 @@ widget_dependencies <- function(name, package){
 # Generates a <script type="application/json"> tag with the JSON-encoded data,
 # to be picked up by htmlwidgets.js for static rendering.
 widget_data <- function(x, id, ...){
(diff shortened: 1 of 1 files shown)


R/htmlwidgets.R
```diff
@@ -269,17 +274,25 @@ shinyRenderWidget <- function(expr, outputFunction, env, quoted) {
...
     }
-    x <- .subset2(instance, "x")
     deps <- .subset2(instance, "dependencies")
     deps <- lapply(
       htmltools::resolveDependencies(deps),
       shiny::createWebDependency
     )
-    evals = JSEvals(x)
-    list(x = x, evals = evals, deps = deps)
+    payload = modifyList(createPayload(instance), list(deps = deps))
```
yihui: `c(createPayload(instance), list(deps = deps))` looks more natural to me here compared to `modifyList()`.

R/htmlwidgets.R
```diff
@@ -269,17 +274,25 @@ shinyRenderWidget <- function(expr, outputFunction, env, quoted) {
...
   }
 
   # mark it with the output function so we can use it in Rmd files
   shiny::markRenderFunction(outputFunction, renderFunc)
 }
 
+# Helper function to create payload
+createPayload <- function(instance){
+  if (!is.null(instance$preRenderHook)){
+    instance <- instance$preRenderHook(instance)
```
yihui: Do we want `instance$preRenderHook <- NULL` here? Because I don't see the value of serializing the R function into JSON.

R/htmlwidgets.R
ramnathv: Yeah. That makes sense. I am just so used to using `modifyList`. I will make the change.

R/htmlwidgets.R
ramnathv: Yeah. That makes sense. So if I understand correctly, the logic is that once you apply the pre render hook, and we have created the payload, we no longer need to carry it. Right?

yihui: That is right. I'll leave it to you guys to decide whether to get rid of this function here. I only read the latter half of the discussion, so I'm not the best person to decide whether this makes sense.

ramnathv: I think it makes a lot of sense @yihui. I can't see any reasons for the hook function to be carried along, once it has been applied, since there is no use for it in the json.

hafen:
Suppose the `preRenderHook` modifies the width and height of the widget.  Would you want to update the sizingPolicy?

timelyportfolio:
I like this.  Glad we had some good use cases to broach the topic.  Similarly, I've been trying to think through this on the JS side.

ramnathv:
@hafen If you explicitly set `width` and `height`, it will overrule what you specify in the `sizingPolicy`. Do you have a use case where you will have to meddle with the `sizingPolicy` pre-render?

@timelyportfolio I have been doing some preliminary work on JS hooks (pre and post render), and will post something on that soon.

timelyportfolio:
forgot started discussion of JS hooks in https://github.com/ramnathv/htmlwidgets/issues/65

hafen:
@ramnathv, I was playing around with some `preRenderHook` functions and found that explicitly setting `width` and `height` inside the hook is ignored.  Looking at the code this is because `resolveSizing` is called before the hook is called.  At the time the hook is called, sizing decisions have already been made.

I actually don't have a need or use case for this at the moment - just curious if you want to allow the hook to modify the dimensions.  Probably not important right now.

ramnathv:
Good point @hafen. I would have to think through this one.

hafen:
@ramnathv - maybe we don't need to address it until there is a valid use case for it.  I stumbled on it by accident when I put something in my `preRenderHook` that modified the width/height when I should have been doing it further upstream before the hook is called.  I'm all for getting this feature merged as soon as we can :).

ramnathv:
Sounds good @hafen. Once @jcheng5 and @jjallaire have reviewed this PR, it is good to be merged.

jcheng5:
I'll take a look tomorrow. Thanks @hafen.

jjallaire:
Looks good to me!

R/htmlwidgets.R
jcheng5: The `instance` is just a list, right? So we're not modifying the original widget instance, just the temporary copy that's about to be rendered to JSON?

jcheng5:
Thanks @hafen!

R/htmlwidgets.R
ramnathv: Yes. `instance` is a list representation of the widget object, which gets modified just prior to conversion to JSON.

R/htmlwidgets.R
```diff
@@ -269,17 +274,26 @@ shinyRenderWidget <- function(expr, outputFunction, env, quoted) {
...
     }
-    x <- .subset2(instance, "x")
     deps <- .subset2(instance, "dependencies")
     deps <- lapply(
       htmltools::resolveDependencies(deps),
       shiny::createWebDependency
     )
-    evals = JSEvals(x)
-    list(x = x, evals = evals, deps = deps)
+    payload = c(createPayload(instance), list(deps = deps))
```
yihui: @ramnathv Sorry I missed a minor problem here: there is no need to assign the list to `payload` here. It does not really hurt since `=` returns the RHS invisibly, so the returned value is still correct.

ramnathv: @yihui That is right. I will leave as-is for now, since it does not create any problems.

yihui: I prefer removing it, though :)

PR #96, opened 2015-04-13
Some cleanup

yihui:
@ramnathv I don't think we really need `to_json()` now. Not sure if you agree.


ramnathv:
I agree that it can be taken out now. I am assuming none of the packages other than mine made use of it anyway. @timelyportfolio just wanted to give you the heads up that we are removing `to_json` from `htmlwidgets` just in case you are using it.

PR #104, opened 2015-05-01
Fix a bug with </script> appearing in widget data

jcheng5:
See comment in the commit.

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -122,8 +122,15 @@ widget_dependencies <- function(name, package){
 # Generates a <script type="application/json"> tag with the JSON-encoded data,
 # to be picked up by htmlwidgets.js for static rendering.
 widget_data <- function(x, id, ...){
+  # It's illegal for </script> to appear inside of a script tag, even if it's
+  # inside a quoted string. Fortunately we know that in JSON, the only place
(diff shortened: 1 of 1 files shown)


ramnathv:
@jcheng5 I presume this fixes a security bug?

jcheng5:
No, I don't think it's security related--good guess though! If `</script>` appears in the widget data, it'll simply break that data's ability to load.

I'm pulling this out of the air, forgive any mistakes:

```html
<script type="application/json" for="widget-1234">
{
  width: null,
  height: null,
  x: {
    title: "Why </script> is problematic"
  }
}
</script>
```

When widget-1234 goes to load/parse its JSON, it'll fail, since the browser will consider the script block ended right after `title: "Why `.

ramnathv:
Got it. I will test tonight to make sure that it doesn't break anything and merge.

yihui:
FWIW, Joe has used this in the leaflet docs: http://rstudio.github.io/leaflet/ I don't see anything wrong there (look for `\u003c` in the HTML source). So I guess this PR should be good to go. It is a little unfortunate we didn't merge it for the 0.4 release.

ramnathv:
Sorry for missing this one.

yihui:
Thanks. Can you also bump the version?

ramnathv:
Done! Thanks

Issue #102, opened 2015-04-28
dynamically adjust `height` and `width` of a `htmlwidget` in Shiny


timelyportfolio:
https://github.com/rich-iannone/DiagrammeR/issues/93 brings up an interesting point that I have not considered before.  What is the best way to handle this, since `initialize` and `resize` are not called in this scenario?  I can only think to provide `height` and `width` also in `x` so that the adjustment can be made in `renderValue`.

jrowen:
I'm also running into this issue with `rhandsontable` and `shinydashboard`,  
```
shiny::runGitHub("rhandsontable", "jrowen", 
             subdir="inst/examples/rhandsontable_dash")
```
When resizing the window and show/hiding the sidebar, I see the boxes resize but not the widget.  As noted, `resize` and `initialize` are not called.  Was a solution to this issue ever identified?

ziyadsaeed:
This is a problem with DT package as well. Im using the responsive plugin in DT to have the table resize the width.

PR #106, opened 2015-05-12
Call widget.resize in more situations

jcheng5:
e.g. rstudio/leaflet wasn't working if you had a widget like this:

```{r}
library(leaflet)
leaflet() %>% addTiles()
```

in an rmarkdown ioslide on the 5th slide and the slide deck was
showing slide 1 when the browser loaded it.

I'll submit more repro cases in the pull request.

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -452,7 +452,7 @@
 
         if (binding.resize) {
           var lastSize = {};
-          on(window, "resize", function(e) {
+          var resizeHandler = function(e) {
             var size = {
               w: sizeObj ? sizeObj.getWidth() : el.offsetWidth,
               h: sizeObj ? sizeObj.getHeight() : el.offsetHeight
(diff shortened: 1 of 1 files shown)


jjallaire:
This looks good to me. @ramnathv ?

ramnathv:
This looks good to me. Although this makes me wonder, if we need and can build a more generic solution that will allow resizing to be triggered by other frameworks/widgets. Ideally, `htmlwidgets` should provide a mechanism that `ioslides` can take advantage of to postpone rendering till the particular slide is displayed. We can merge this in, and think about that for the future. @jcheng5 @jjallaire any thoughts on this?

jcheng5:
I think the hooks are there already, if you don't want to render when you're not visible you can have your widget's `renderValue` callback not do anything when the element's offsetWidth/offsetHeight are 0. And then have your `resize` callback call `renderValue`, as in previous examples we've looked at.

However, we still need to make sure `resize` gets called at the appropriate time, which wasn't the case with ioslides (at least not in all scenarios) without this PR. One idea JJ and I batted around earlier today was triggering the window's resize event from ioslides. The shown/hidden events that are subscribed in this PR would also work, but those depend on jQuery to work cross platform AFAICT.

jcheng5:
Sorry for not delivering the promised examples yet. I am seeing one weird behavior that may be due to just a weird version of Shiny installed, but I want to make sure.

jcheng5:
Two examples:
https://gist.github.com/jcheng5/aff6ef5dbb76bb140f30

A Leaflet bugfix is also needed to fix this bug. I've just merged it into master, so `devtools::install_github('rstudio/leaflet')` should do it.

The "Shiny map" slide should render fine even without this fix; the others do need the fix.

To repro the problem, make sure when the browser loads the slide deck, that it's starting at slide 1.

ramnathv:
Thanks @jcheng5. I won't be able to test this till late next week. But I am good with merging this since you and @jjallaire have tested it, and it only modifies behavior specific to `ioslides` minimizing the chances of any regression.

timelyportfolio:
possibly related to #102; +1 for more resize

PR #111, opened 2015-05-18
Check shiny version

yihui:
I just discussed this with @jcheng5 and @wch, and we thought it'd be a good idea to check the version of shiny in htmlwidgets (and vice versa https://github.com/rstudio/shiny/pull/830), in case users upgraded one package but not the other.

Checking takes place in both `.onLoad()` and `shinyWidgetOutput()`/`shinyRenderWidget()`, so hopefully users should have good chances to see this message.

### DIFF
--- a/R/knitr-methods.R
+++ b/R/knitr-methods.R
@@ -20,6 +20,9 @@ registerMethods <- function(methods) {
 }
 
 .onLoad <- function(...) {
+  # warn if the version of shiny is lower than what was specified in DESCRIPTION
+  checkShinyVersion(error = FALSE)
+
   # htmlwidgets provides methods for knitr::knit_print, but knitr isn't a Depends or
   # Imports of htmltools, only an Enhances. Therefore, the NAMESPACE file has to
   # declare it as an export, not an S3method. That means that R will only know to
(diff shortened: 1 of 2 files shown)


R/knitr-methods.R
```diff
@@ -20,6 +20,9 @@ registerMethods <- function(methods) {
 }
 
 .onLoad <- function(...) {
+  # warn if the version of shiny is lower than what was specified in DESCRIPTION
+  checkShinyVersion(error = FALSE)
```
wch: I think this will raise a NOTE with R CMD check, because it's in `.onLoad` instead of `.onAttach`.

yihui:
Just fixed a minor issue in this PR, and it should be ready to go now. Cc @jjallaire

R/knitr-methods.R
wch: On the other hand, the note probably won't get raised on the CRAN check machines, because they will have the new version of Shiny on there before htmlwidgets is submitted.

yihui: R CMD check is not that smart :) It cannot detect `packageStartupMessage` when it is called from another function.

yihui: Oh that is right.

jjallaire:
This isn't working as I expect for me. Here's a transcript of my session after building/installing htmlwidgets master:

```
> packageVersion("shiny")
[1] ‘0.11.1.9003’
> library(htmlwidgets)
> 
```

Shouldn't loading the library issue a warning?

yihui:
@jjallaire Sorry, will fix it in a minute.

PR #114, opened 2015-05-21
Fix bug where resize() isn't called on some widgets

jcheng5:
If there were multiple static widgets in the same document, not
all would receive resize() calls. Repro case is to create an
R Markdown ioslides presentation, and on slide 4 add one leaflet
map, and on slide 5 add another one. Only the second one would
be rendered properly. Here's an example, see slide 4:

https://rawgit.com/ldecicco-USGS/toxEval/master/samplePres.html

This was due to me using function-scoped variables inside of a
loop that created closures:

```javascript
for (var i = 0; i < bindings.length; i++) {
  var binding = bindings[i];
  var matches = binding.find(document.documentElement);
  for (var j = 0; j < matches.length; j++) {
    var el = matches[j];
    function resizeHandler() {
      // do stuff with binding and el
    }
  }
}
```

The binding and el variables are scoped to the entire function,
so all the iterations all share the same variable, and therefore
all the resizeHandler closures are bound to the same binding and
el values (the final ones).

The solution is to change the for-loops into forEach calls, since
the callback function into forEach will result in an isolated
scope being created for each iteration.

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -49,6 +49,17 @@
     return target;
   }
 
+  // IE8 doesn't support Array.forEach.
+  function forEach(values, callback, thisArg) {
+    if (values.forEach) {
+      values.forEach(callback, thisArg);
+    } else {
+      for (var i = 0; i < values.length; i++) {
+        callback.call(thisArg, values[i], i, values);
+      }
+    }
+  }
+
   // Replaces the specified method with the return value of funcSource.
   //
   // Note that funcSource should not BE the new method, it should be a function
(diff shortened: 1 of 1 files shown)


jcheng5:
Nothing like releasing to CRAN to ensure that an important bug will be discovered and fixed...

jjallaire:
Could you add a NEWS entry for this fix? In the meantime I'll merge it so it's easier to install for users that need the fix.

jjallaire:
Maybe bump the version as well.

jcheng5:
OK, I'll make the NEWS/DESCRIPTION updates when I get back from lunch.

ramnathv:
Thanks @jcheng5. Can you delete this branch if it is not needed anymore?

PR #115, opened 2015-05-23
Only replace </script>, instead of all HTML tags

yihui:
This PR narrows the fix in #104. If it is unnecessary to uglify all HTML tags, I'd vote not to do so. Will this work? @jcheng5

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -129,9 +129,9 @@ widget_data <- function(x, id, ...){
   # repro for the bug this gsub fixes is to have the string "</script>" appear
   # anywhere in the data/metadata of a widget--you will get a syntax error
   # instead of a properly rendered widget.
-  tags$script(type="application/json", `data-for` = id,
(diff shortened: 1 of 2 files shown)


R/htmlwidgets.R
```diff
@@ -130,7 +130,7 @@ widget_data <- function(x, id, ...){
   # anywhere in the data/metadata of a widget--you will get a syntax error
   # instead of a properly rendered widget.
   tags$script(type="application/json", `data-for` = id,
-    HTML(gsub("<", "\\\\u003c", toJSON(createPayload(x))))
+    HTML(gsub("</script>", "\\\\u003c/script>", toJSON(createPayload(x))))
```
jcheng5: I'd set `ignore.case=TRUE` and match on `</script`, not `</script>`. (Test case: `</SCRIPT >`

yihui: Good point. Done.

jjallaire:
I'm okay with this but want to have @jcheng5 merge it since it's a refinement of his code.

PR #117, opened 2015-05-25
Update website

yihui:
This PR may look a little overwhelming. If you want to review it, I recommend you to go to https://github.com/yihui/htmlwidgets/commits/gh-pages, open the last commit, then press `P` to navigate through the commits. There are two big commits that you can ignore:

- https://github.com/yihui/htmlwidgets/commit/0f0a91825517a7fc44c80527610dbdc5397bf071 removed the jQuery source files; I believe they were introduced by the metricsgraphics package, and they are useless (we can just use `jquery/jquery.min.js`)
- https://github.com/yihui/htmlwidgets/commit/9c367e85f518af49c37047c25f21421c59feca3a contains changes due to upgrading R packages

I added a `compile.R` that merges different versions of the same library by discarding lower versions. Version numbers in the directory names are stripped off to make it easier to know what exactly have changed the next time we update the website, otherwise we will keep on introducing higher versions of libraries into this branch (e.g. `datatables-binding-0.0.1`, `datatables-binding-0.0.2`, ... and lower versions are actually not used any more). This is a strategy I have been using on the DT website, and also introduced it to the leaflet website. This makes me feel more comfortable and safer since the changes in GIT are often clear enough so I know I didn't introduce random unexpected changes hidden behind a giant diff. For the same reason, I set `options(htmlwidgets.TOJSON_ARGS = list(pretty = TRUE))` to make diffs in JSON more human-readable.

I have checked all pages, and all widgets still work well.


jjallaire:
Thanks! These are great changes (I've never been comfortable with the way the lib directory worked either). Also glad to see the wart of out of sync docs removed

ramnathv:
Thanks @yihui. This is solid.

PR #122, opened 2015-06-01
Pass the shiny session object and output id to preRenderHook()

yihui:
This change will significantly simplify the usage of `DT::datatable(server = TRUE)`. Basically I passed two objects `shinysession` (the shiny session object) and `name` (the output id) to the hook `preRenderHook()`. I need them to register a data object in a shiny session so DataTables can send Ajax requests to the URL returned by the registration. Previously, the server-side DataTables looks like this:

``` r
shinyServer(function(input, output, session) {
  action = dataTableAjax(session, iris)
  widget = datatable(iris, server = TRUE, options = list(
    ajax = list(url = action)
  ))
  output$tbl = DT::renderDataTable(widget)
})
```

With this change, users will no longer have to worry about DataTables internals like the Ajax URL. Just `server = TRUE`, and it works.

``` r
shinyServer(function(input, output, session) {
  widget = datatable(iris, server = TRUE)
  output$tbl = DT::renderDataTable(widget)
})
```

I have taken backward compatibility into consideration: if the hook only has one argument, I will ignore the additional arguments `shinysession` and `name`.

Hopefully this is a simple enough PR that can get merged quickly. The corresponding change in DT is here: https://github.com/rstudio/DT/commit/ba899e37f0725ed89c179f753b60dc22466d54e7 I'm thinking about the CRAN release of DT, and will need a new release of htmlwidgets accordingly. Thanks!

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -271,7 +271,7 @@ shinyRenderWidget <- function(expr, outputFunction, env, quoted) {
   func <- shiny::exprToFunction(expr, env, quoted)
 
   # create the render function
-  renderFunc <- function() {
+  renderFunc <- function(shinysession, name) {
     instance <- func()
     if (!is.null(instance$elementId)) {
       warning("Ignoring explicitly provided widget ID \"",
@@ -283,7 +283,7 @@ shinyRenderWidget <- function(expr, outputFunction, env, quoted) {
       htmltools::resolveDependencies(deps),
       shiny::createWebDependency
     )
-    payload <- c(createPayload(instance), list(deps = deps))
+    payload <- c(createPayload(instance, shinysession, name), list(deps = deps))
     toJSON(payload)
   }
 
@@ -303,11 +303,14 @@ checkShinyVersion <- function(error = TRUE) {
 }
 
 # Helper function to create payload
-createPayload <- function(instance){
-  if (!is.null(instance$preRenderHook)){
-    instance <- instance$preRenderHook(instance)
-    instance$preRenderHook <- NULL
+createPayload <- function(instance, ...){
+  hook <- instance$preRenderHook
+  if (is.function(hook)) {
+    instance <- if (length(list(...)) && length(formals(hook)) > 1) {
+      hook(instance, ...)
+    } else hook(instance)
   }
+  instance$preRenderHook <- NULL
   x <- .subset2(instance, "x")
   list(x = x, evals = JSEvals(x))
 }
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -306,7 +306,7 @@ checkShinyVersion <- function(error = TRUE) {
 createPayload <- function(instance, ...){
   hook <- instance$preRenderHook
   if (is.function(hook)) {
-    instance <- if (length(list(...)) && length(formals(hook)) > 1) {
+    instance <- if (length(formals(hook)) > 1 && length(list(...))) {
       hook(instance, ...)
     } else hook(instance)
   }


jjallaire:
At first look this seems fine to me but @jcheng5 and @ramnathv know these internals better than me so should also take a closer look. Assuming those reviews go well we can merge them right away. In terms of CRAN release we are about 2 weeks post our last CRAN release so plan on doing another one in ~ 2 weeks (to get this plus the important resize fix that Joe put it).

ramnathv:
@yihui I will be able to take a closer look at this on Wed. Hope that works well with your timeline for DT.

jcheng5:
@ramnathv please don't merge until I've had a chance to take a close look as well, thanks.

jcheng5:
@yihui we have an incentive to make this feature work without depending on a new version of htmlwidgets, since htmlwidgets just went to CRAN very recently. We can certainly get at the session without htmlwidgets' help, using `shiny::getDefaultReactiveDomain`. However I'll have to think more about how to get at the name, or maybe we can work around it.

ramnathv:
@jcheng5 I won't merge before your okay. All I was saying was that I will be able to get to this only on Wed :)

jcheng5:
@ramnathv Yep, understood, thanks :)

yihui:
We used a hackish workaround in DT and released it to CRAN yesterday. I hope this PR can be merged anyway. No need to release a new version of htmlwidgets to CRAN soon, though.

jcheng5:
So if you want to opt into receiving this additional info, you would add `...` to your `preRenderHook` function? And I guess you would need to ignore the `...` for non-Shiny contexts?

yihui:
Yes, package authors can check `length(list(...))` to decide what to do if the hook function has to do different things in Shiny and non-Shiny contexts.

jcheng5:
Ugh, yuck... that's both unintuitive and totally precludes us from adding new non-Shiny arguments later on.

ramnathv:
I would agree with @jcheng5 on this. Let us think through this to see if we can come up with a solution that does not preclude future additions.

jcheng5:
We might also wait until we see another user ask for this... I don't know if this is one we want to preemptively solve, given that we already found a solution that works well, inelegant as it may be.

ramnathv:
So are you suggesting we merge this change in @jcheng5 since it works?

jcheng5:
No, the opposite--I'm suggesting that we close this PR until it comes up again. (In my previous comment, I meant that we worked around the need for this feature in DT.)

ramnathv:
Ah got it. Thanks.

yihui:
@jcheng5 It is a little hard to give up when I feel a simple solution is so close, and we have to give up because of the potential future problems (which makes sense but we don't know if such problems can be real or not). Well, how about using a Shiny-specific hook instead of abusing `preRenderHook`? e.g. `preShinyHook`?

jcheng5:
My philosophy is that when designing APIs that are intended to be consumed by other library developers, coming up with the solution is *by far* the easy part. The hard part is designing features so that they are orthogonal to each other and do not box you in from growth.

Hmmm... a different way we could go is to add a function to Shiny that tells you what the name of the currently-rendering output is (in the same way that we have `getDefaultReactiveContext()` to retrieve the current session). The fact that we already do this for the session makes me less hesitant to take this approach. What do you think?

yihui:
Thanks! That will work, too, as long as `shiny::getDefaultReactiveDomain()` is always the same thing as `shinysession`, and `shiny::getDefaultReactiveContext()` is the same as `name` here in the `renderFunc()`.

Yep, I understand the point of considering future extensibility. It is just that in this particular case, I feel a new hook function for Shiny should be reasonably safe.

jcheng5:
Just a note to future-self that any consideration of this feature will now have to consider Shiny modules/namespaces as well.

PR #130, opened 2015-06-28
Increase pandoc stack size to 64MB for saveWidget

jjallaire:
This is often required in order for widgets with large base64 payloads to render into standalone HTML (several reports of this with embedded leaflet maps have already been made).

Note that htmlwidgets includes a clone of the pandoc function from rmarkdown (so as to not have to depend on rmarkdown). I'll make the same change in rmarkdown, and look to factor those functions out of rmarkdown and into a pandoc package to eliminate the duplication.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: htmlwidgets
 Type: Package
 Title: HTML Widgets for R
-Version: 0.5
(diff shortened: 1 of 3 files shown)


jcheng5:
LGTM

ramnathv:
It looks good. I am merging it in.

PR #131, opened 2015-06-29
set pandoc stack size to 512mb (same as GHCi) and allow override

jjallaire:
The user I put this fix in for actually had another leaflet map that blew past the 64MB (I tried 128MB and it worked fine). Note that the size of the file generated by saveWidget was only ~ 750kb so this wasn't an edge case that we shouldn't expect to see crop up again.

I did some research into GHC stack sizes and found that while the default for compiled programs was 8MB, GHCi (their interactive front-end) uses a default of 512MB. Furthermore, it sounds like this just establishes a maximum stack size rather than pre-allocating the pages:

http://stackoverflow.com/questions/20312960/non-tail-recursive-function-not-blowing-up-in-ghci-why/20313059#20313059

Given this I see no reason not to set our stack size to 512MB (as this PR does). However, at the same time I expose the knobs for users to tune it up and down. However, perhaps it's enough to just use 512MB and be done with it. 

Interested in what others think about (a) what the right default limit is; and (b) whether we should make this tunable or not.

### DIFF
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -14,7 +14,7 @@ pandoc_available <- function(version = NULL) {
     FALSE
 }
 
-pandoc_self_contained_html <- function(input, output) {
+pandoc_self_contained_html <- function(input, output, stack_size = "512m") {
 
   # make input file path absolute
   input <- normalizePath(input)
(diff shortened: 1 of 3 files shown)


jjallaire:
Closed to pursue a different approach: https://github.com/ramnathv/htmlwidgets/pull/132

PR #132, opened 2015-06-29
Increase pandoc stack size to 512 and allow override via option

jjallaire:
Here's another approach to same issue addressed in https://github.com/ramnathv/htmlwidgets/pull/131 (note that only one not both of these PRs should ultimately be merged).

I think I like the global option better because (a) It doesn't clutter up the calling interface of the pandoc functions; (b) It allows user control over the behavior even when developers haven't expressly exposed a knob to tune it; (c) All packages using pandoc can benefit from a single expression of user intent regarding stack usage.

I'm going to close the first PR for now and we can re-open it if folks prefer the first approach.

### DIFF
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -36,8 +36,7 @@ pandoc_self_contained_html <- function(input, output) {
     output = output,
     options = c(
       "--self-contained",
-      "--template", template,
-      "+RTS", "-K64m", "-RTS"
+      "--template", template
     )
   )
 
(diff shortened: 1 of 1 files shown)


ramnathv:
@jjallaire I prefer this approach too. I don't see any drawbacks associated with setting this as a global option.

jjallaire:
Okay, I'll go ahead and merge this. It seems like there's no harm in setting a relatively high stack size given that GHC will still grow it's stack dynamically. @jcheng5 Let us know if you have a different perspective on what the default value should be.

PR #154, opened 2015-09-05
add as.iframe to embed widgets as iframe

ramnathv:
This PR allows html widgets to be embedded as iframes. It works by saving the widget to html using saveWidget and then using htmltools to embed it as an iframe.

By default, it saves the widget html to file using selfcontained = TRUE and then uses the srcdoc attribute to embed it as an inline iframe. This allows the resulting html document to be selfcontained. The srcdoc attribute is not supported by IE, and for some reason does not appear to be supported in the Rmd previewer as well.

I have added notes to make it clear that

1. This only works for static Rmds.
2. The iframe wont display in the RStudio viewer when selfcontained = TRUE is chosen.


ramnathv:
I am sending another PR, since the previous one ran into merge commits due to updates on the master. It is the same PR, except for notes that make it amply clear that this is to be used only in static Rmds, with the limitation that the iframe won't be viewable in RStudio viewer.

timelyportfolio:
Yes, please :)

yutannihilation:
Hi! This feature seems very nice. I'm looking forward to using this :+1: 

If `as.iframe()` is used only with knitr, how about implementing `as.iframe` option in `knit_print.htmlwidget()` instead?

Here is my sample code (I don't test this yet):

```r
knit_print.htmlwidget <- function (x, ..., options = NULL)
{
  if (!is.null(rmarkdown::metadata$self_contained) &&
      rmarkdown::metadata$self_contained) {
    file <- basename(tempfile(fileext = ".html"))
    selfcontained <- if (is.null(rmarkdown::metadata$self_contained)) TRUE else rmarkdown::metadata$self_contained
    
    htmlwidgets::saveWidget(x, file = file, selfcontained = selfcontained)
    content <- if (selfcontained) {
      on.exit(unlink(file), add = TRUE)
      list(srcdoc = paste(readLines(file), collapse = "\n"), width = options$out.width, height = options$out.height)
    } else {
      list(src = file, width = options$out.width, height = options$out.height)
    }
    
    knitr::knit_print(htmltools::tag("iframe", content), options = options, ...)
    
  } else {
    knitr::knit_print(toHTML(x, standalone = FALSE, knitrOptions = options), options = options, ...)
  }
}
```

`as.iframe` option will be specified in YAML Front-matter.

```yaml
output:
  html_document:
    as.iframe: true
```

timelyportfolio:
@ramnathv, can we get your version merged in `0.5.3`?

ramnathv:
Yes. I will most likely close this PR and initiate a fresh one.

timelyportfolio:
is this ever going to happen?

ramnathv:
This requires some cleanup given the conflicts.

timelyportfolio:
I am happy to fix it up as long as there is an expectation/desire to integrate.  Just let me know.

timelyportfolio:
In general, my new `master` with this merged seems to work fine however.

timelyportfolio:
@ramnathv circling back on this, did this get done?  I don't remember and couldn't find any new implementations.

PR #155, opened 2015-09-06
Fix #151: In JS bindings, resize() can't call renderValue()

### LINKED ISSUE #151: In JS bindings, resize() can't call renderValue()
For bindings that want to re-render on resize, we'd like this to work:

```r
renderValue: function(el, x, instance) {
  instance.lastValue = x;
  // render...
},
resize: function(el, width, height, instance) {
  this.renderValue(el, instance.lastValue, instance);
},
```

This doesn't work today because `this.renderValue()` doesn't call the `renderValue` right in front of you, but a monkeypatched version that htmlwidgets inserts on your behalf when you register the widget. The monkeypatched version expects not an `x` value but `data`, from which it retrieves `x`:

https://github.com/ramnathv/htmlwidgets/blob/a8276d4378e5bb8603b24a9b7aac617bed2b11c4/inst/www/htmlwidgets.js#L413

This should be fixed by not using the monkeypatching approach for Shiny bindings, but rather, decorator. In other words, the Shiny binding should be a wrapper around the htmlwidget binding, instead of taking the htmlwidget binding and monkeypatching it into a Shiny binding. (Boy, it seems pretty obvious when I put it that way...)

jcheng5:
This commit ensures that calling `this.renderValue()` from inside a widget method like `resize()` calls the actual `renderValue` code that's in the widget, not a Shiny-friendly wrapper version that we inject from htmlwidgets. Rather than overriding these methods, we now create a separate ShinyOutput object that merely decorates the definition.

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -75,6 +75,45 @@
     target[methodName] = funcSource(superFuncBound);
   }
 
+  // Add a method to delegator that, when invoked, calls
+  // delegatee.methodName. If there is no such method on
+  // the delegatee, but there was one on delegator before
+  // delegateMethod was called, then the original version
+  // is invoked instead.
+  // For example:
+  //
+  // var a = {
+  //   method1: function() { console.log('a1'); }
+  //   method2: function() { console.log('a2'); }
+  // };
+  // var b = {
+  //   method1: function() { console.log('b1'); }
+  // };
+  // delegateMethod(a, b, "method1");
+  // delegateMethod(a, b, "method2");
+  // a.method1();
+  // a.method2();
+  //
+  // The output would be "b1", "a2".
+  function delegateMethod(delegator, delegatee, methodName) {
+    var inherited = delegator[methodName];
+    delegator[methodName] = function() {
+      var target = delegatee;
+      var method = delegatee[methodName];
+
+      // The method doesn't exist on the delegatee. Instead,
+      // call the method on the delegator, if it exists.
+      if (!method) {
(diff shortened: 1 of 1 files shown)


jcheng5:
I thought of a different way to do this.

jcheng5:
The "different way to do this" didn't pan out. The idea was to not override+delegate `renderValue`, but rather to override `onValueChange`, a lower-level method on `Shiny.OutputBinding`. In other words, instead of intercepting the call to `renderValue(el, data)` and proxying that through to the htmlwidget version (`renderValue(el, data, instance)`), we act a step sooner and intercept the call to `onValueChange`, which is what normally calls `renderValue(el, data)`, and have it instead call `renderValue(el, data, instance)`.

That approach failed because there's also the `resize` method, which has no equivalent `onValueChange` low-level call.

yihui:
So far I guess I have received at least three bug reports related to this. One from Zev Ross at the Shiny Dev Conference, and the other two at https://github.com/rstudio/DT/issues/238 and https://github.com/rstudio/shiny/issues/1102 So I really hope we can fix it.

jcheng5:
@yihui I didn't realize DT used this (non-working) pattern. This PR is ready to go other than the merge conflicts since I wrote it. I can get it merged on Monday.

jjallaire:
Let's merge it Monday then (since a CRAN release is imminent)

yihui:
@jcheng5 I resolved the conflict (appears to be simple) and opened #189. Not sure if that will actually make it faster or even slower for your fix to be merged :)

PR #169, opened 2015-11-13
Fix escaping of HTML tags in JSON strings. Closes #168

### LINKED ISSUE #168: Some HTML strings in JSON result in mangled conversion
For example, if you view this profvis htmlwidget with the default print method, it comes up fine:

```R
# devtools::install_github("rstudio/profvis")
library(profvis)

p <- profvis({
  pause(0.1)
  #</body></html> &&
})

p
```

But if you save it with `savewidget`, the browser will have an error.

```R
htmlwidgets::saveWidget(p, "p.html")
```

In Chrome, it says:
> Uncaught SyntaxError: Unexpected token &

The problem is that the JSON gets mangled. With the printed version, the HTML document contains this in the JSON:

```
"content":"p <- profvis({\n  pause(0.1)\n  #</body></html> &&\n})\n"
```

In the saved version, the escaping switches to HTML escaping after it sees `</body></html>`:

```
"content":"p <- profvis({\n  pause(0.1)\n  #</body></html>
<p>&amp;&amp;\n})\n&quot;
```

Additionally, if the code is re-run with strings like `</body> foo </html>`, or just `</body>` or `</html>`, the problem doesn't happen. For example, this works:

```R
p <- profvis({
  pause(0.1)
  #'</body> &&
})
htmlwidgets::saveWidget(p, "p.html")
```

One solution that @jcheng5 and I discussed is to replace all instances of `</` with `\u003c/`.

cc @yihui

wch:
This fixes #168.

The previous code used the replacement string `\\\\u003c`, but I believe this is incorrect -- it should be `\\u003c`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: htmlwidgets
 Type: Package
 Title: HTML Widgets for R
-Version: 0.5.1
+Version: 0.5.2
 Date: 2015-06-21
 Authors@R: c(
     person("Ramnath", "Vaidyanathan", role = c("aut", "cph")),
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,8 @@
-htmlwidgets 0.5.1 (unreleased)
+htmlwidgets 0.5.2 (unreleased)
 -----------------------------------------------------------------------
+
+* Fix a bug where the string "</body></html>" in the widget data caused
+  `saveWidget()` to have malformed output. (#168)
 
 * Increase pandoc stack size to 512M for saveWidget (often required for
   e.g. larger embedded leaflet maps). Stack size can also be controlled
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -129,8 +129,13 @@ widget_data <- function(x, id, ...){
   # repro for the bug this gsub fixes is to have the string "</script>" appear
   # anywhere in the data/metadata of a widget--you will get a syntax error
   # instead of a properly rendered widget.
+  #
+  # Another issue is that if </body></html> appears inside a quoted string,
+  # then when pandoc coverts it with --self-contained, the escaping gets messed
+  # up. There may be other patterns that trigger this behavior, so to be safe
+  # we can replace all instances of "</" with "\\u003c/".
   payload <- toJSON(createPayload(x))
-  payload <- gsub("</(script)>", "\\\\u003c/\\1>", payload, ignore.case = TRUE)
+  payload <- gsub("</", "\\u003c/", payload, fixed = TRUE)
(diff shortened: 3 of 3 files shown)


jjallaire:
This looks good to me. @jcheng5 could you review also and merge?

wch:
Should there also be a version bump and NEWS item?

jjallaire:
Yes definitely!

jcheng5:
I just realized... might there be a similar problem with `<!--` (without a matching `-->`)?

jcheng5:
I can't convince myself that it's safe to replace <!--, nor can I get pandoc to choke on it. Let's move forward with this as it's clearly an improvement.

yihui:
This is really a Pandoc issue, right? As I expressed in #115, I don't like all HTML tags to be escaped because it makes the JSON string looks ugly (if we care whether the JSON string is human-readable). I'd rather deal with this issue case by case, e.g. add

```r
gsub("</(body>\\s*</html>)", "\\\\u003c/\\1", payload)
```

I'm fine if the decision was final.

@wch `\\\\u003c` was correct because `fixed = FALSE` in `grep()` by default. `\\\\` means `\\` when it is in the replacement string.

jcheng5:
This is a grey enough area that I think it's better to err on the side of over-escaping for correctness rather than under-escape for source readability. According to the HTML4 spec (emphasis mine):

> Although the STYLE and SCRIPT elements use CDATA for their data model, for these elements, CDATA must be handled differently by user agents. Markup and entities must be treated as raw text and passed to the application as is. **The first occurrence of the character sequence "</" (end-tag open delimiter) is treated as terminating the end of the element's content.** In valid documents, this would be the end tag for the element.

http://www.w3.org/TR/html4/types.html#type-cdata

(HTML5 clears this up by not using CDATA for style and script; instead, there are hardcoded parsing rules for script blocks, so only `</script>` and `<!--` remain problematic.)

wch:
@yihui Oh right, that makes sense for the `\\\\` vs `\\` thing.

Regarding escaping `"</(body>\\s*</html>)"`, that wouldn't fix it in all cases, because in my testing:

* `</body></html>` results in the problem.
* `</body> </html>` results in the problem
* `</body> text </html>` results in the problem.
* `</body> <script></script> </html>` does not result in the problem.

It's not clear exactly which patterns will have a problem when converted by pandoc, so I think it's a good idea to use this more conservative fix. (In addition to the reason that Joe provided.) On the plus side, the pattern `</xxx>` doesn't come up very often in text, so the impact on readability will be limited.

yihui:
Unfortunately it comes up often in **DT**, and that is what bothers me. I'll try to keep my eyes off the JSON object the next time I work on DT :)

PR #171, opened 2015-11-26
Introduce new-style instance-bound widget definitions

jcheng5:
This is part 1 of letting users add arbitrary JS to orchestrate htmlwidget components. Up until now we have mostly focused on inserting a visualization into a page and being done--the only times any code from outside the visualization interacts with that visualization again, is if 1) the container is resized, then we call resize(); and 2) in Shiny contexts, new data can be provided and renderValue() gets called again in that case.

What I would like to allow going forward is arbitrary interaction with htmlwidget components from code outside of the component. You should be able to have your widget expose whatever methods you want, and it should be easy to call those components from JavaScript running elsewhere in the page.

For this to happen we need two things to happen.

1. It needs to be easy and natural for htmlwidget authors to add custom methods that can be invoked on a widget instance. This was not easy nor natural in the past, because the object that you pass into `HTMLWidgets.widget()` was a single object that was responsible for all widget instances of that type. (This also led to an awkward programming style in general.) This PR changes this to one-object-per-widget-instance.
2. You need to be able to easily specify an ID for a widget, and then use that ID from JavaScript to access the widget instance (like `HTMLWidgets.get(id)`). This will be introduced in a separate PR.

Once both items are in place you'll be able to do something like this (assuming appropriate support in the widgets themselves:

```r
widgetGridPage(
  widget1(id = "one", data = iris, ...),
  widget2(id = "two", data = iris, ...),
  tags$script("
    HTMLWidgets.get('one').on('bounds-changed', function(e) {
      HTMLWidgets.get('two').setBounds(e.value);
    });
  ")
)
```

Take a look at the difference between these two otherwise-identical bindings to see the contrast between the existing class-bound definitions and the new instance-bound definitions.

[Class-bound](https://github.com/jcheng5/d3scatter/blob/94e512f6edc15a4ac122e61136e1d40ca6ff8a6e/inst/htmlwidgets/d3scatter.js)
[Instance-bound](https://github.com/jcheng5/d3scatter/blob/470c571f48e6fe895bc2af4bc3cc159d8df63cbe/inst/htmlwidgets/d3scatter.js)

This particular instance-bound example doesn't do anything besides support `renderValue` and `resize`, but you could imagine any number of properties and methods added to that object.

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -647,7 +647,7 @@
   // to rewrite all the logic in this file to accomodate both
   // types of definitions.
   function createLegacyDefinitionAdapter(defn) {
-    return {
+    var result = {
       name: defn.name,
       type: defn.type,
       initialize: function(el, width, height) {
@@ -660,6 +660,15 @@
         return instance.resize(width, height);
       }
     };
+
+    if (defn.find)
+      result.find = defn.find;
+    if (defn.renderError)
+      result.renderError = defn.renderError;
+    if (defn.clearError)
+      result.clearError = defn.clearError;
+
+    return result;
   }
 })();
 
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -476,6 +476,7 @@
             sizeObj ? sizeObj.getWidth() : el.offsetWidth,
             sizeObj ? sizeObj.getHeight() : el.offsetHeight
           );
+          elementData(el, "init_result", initResult);
         }
 
         if (binding.resize) {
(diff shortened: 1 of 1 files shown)


jjallaire:
I think the instance bound definitions will be much easier for developers to get their head around! @ramnathv This looks good to me, let us know your thoughts.

ramnathv:
Thanks @jcheng5. @jjallaire I will go over this tomorrow and add any additional comments/thoughts I have.

ramnathv:
This looks good @jcheng5. Once we have the PR implementing `HTMLwidgets.get`, it would be easier to test the factory approach to defining widgets and ensuring that it will serve the different use cases that we are looking to support. 

I am good with merging this PR in, but maybe it might work better if we consider it along with the next one on `HTMLwidgets.get`. Thoughts?

jcheng5:
Added these functions:

- `HTMLWidgets.getInstance(el)` - Returns the instance associated with an element
- `HTMLWidgets.find([scope], selector)` - Returns the first instance that matches the selector
- `HTMLWidgets.findAll([scope], selector)` - Returns all instances that match the selector

Examples:

```javascript
var foo = HTMLWidgets.getInstance(document.querySelector("#foo"));
var bar = HTMLWidgets.find(containerDiv, "#bar");
var maps = HTMLWidgets.findAll(".leaflet");
```

timelyportfolio:
Should we bump the version?  I have just built my first using the new instance-bound syntax, and I want to require this version of `htmlwidgets` to insure that it will work.

jcheng5:
Sure, go for it.

PR #172, opened 2015-11-30
JavaScript conveniences

jcheng5:
In #171 I added features to make widget instances more easy to manipulate from JavaScript. However, it's not obvious how you actually get your custom JavaScript code onto a web page alongside a widget (assuming you're doing a standalone render of a widget, not in R Markdown or Shiny).

I think there are actually three related problems we should solve for this release:

1. Inject JavaScript (and other invisible elements like CSS stylesheets) alongside a widget when it's rendered standalone. Don't modify anything about the page layout (for example the standalone (re)sizing rules should apply, which is not generally the case when you combine a widget with other HTML like in R Markdown and Shiny)
2. Put multiple widgets, plus additional JS/CSS, into a single page (without R Markdown or Shiny).
3. Custom JavaScript that wants to access widget instance objects need to defer execution until after the widgets have actually been rendered; if they run at page load time, the page will only contain `<script type="application/json">` and empty `<div>`s.

This PR addresses 1 (via `prependContent` and `appendContent`) and 3 (`onStaticRenderComplete`). (I think 2 will be a bigger project and should go into a different PR.)

Here's an example of how you might use the two together.

```r
library(htmlwidgets)
library(leaflet)

leaflet() %>% addTiles() %>% appendContent(
  onStaticRenderComplete("
    var map = HTMLWidgets.find('.leaflet');
    map.on('click', function(e) {
      alert(e.latlng);
    });
  ")
)
```

The `prependContent` and `appendContent` functions insert HTML between the `#htmlwidget_container` and the widget element itself.

```html
<div id="htmlwidget_container">
  <!-- prepended content goes here -->
  <div class="leaflet"> ... </div>
  <!-- appended content goes here -->
</div>
```

The content can be any htmltools objects (`tags`, escaped text, unescaped HTML, lists, and `NULL`) and will simply be stored on the widget itself as additional list elements (`x$prependedContent` and `x$appendedContent` or whatever). If the prepended/appended content has HTML dependencies, those will be correctly handled.


jcheng5:
I got an example from @timelyportfolio showing how he handles this:

https://github.com/hrbrmstr/taucharts/blob/2cc90a4ead0a9ceadc1524b67462a91fb85faaf4/R/tasks.r
https://github.com/hrbrmstr/taucharts/blob/2cc90a4ead0a9ceadc1524b67462a91fb85faaf4/inst/htmlwidgets/taucharts.js#L100-L114

This is quite nice, better than the append/prepend mechanism for most scenarios I think. I'm going to hold off on this PR until I've had a chance to think about this.

ramnathv:
@jcheng5 @timelyportfolio I like the `tasks` mechanism too. The only suggestion I would have is that the callback function signature should include access to `el`, `width`, `height` and `value` (basically what the current version of `renderValue` has). This would allow one to make use of these parameters while specifying the callback.

jcheng5:
I think append/prepend still makes sense for 1) htmlwidgets that don't expose a tasks parameter (especially legacy ones), 2) CSS customizations, and 3) non-intrusively adding additional HTML dependencies. So it looks like the two approaches are at least somewhat complementary.

jcheng5:
OK so scratch "1) htmlwidgets that don't expose a tasks parameter (especially legacy ones)". It's going to work like this:

```r
leaflet() %>% addTiles() %>%
  htmlwidgets::addRenderTask("function(el, x) { this.setZoom(4); }")
```

so all widgets, old and new, will magically gain this ability!

jcheng5:
Anyone have a problem with the function name `addRenderHook`? Hook is a slightly ugly-sounding word to me, but I thought this was more generic than "task" because I could see us expanding this notion in the future:

```r
widget %>%
  addHook("initialize", ...js code...) %>%
  addHook("render", ...js code...) %>%
  addHook("resize", ...js code...)

**Comment by timelyportfolio** [comment]
I have no problems with the word "hook" nor the pull request as changed.  I very much like this ability.

**Comment by ramnathv** [comment]
I feel comfortable with the function name `addRenderHook`. Another option is to use somethin like `on` or `onEvent`.

**Comment by jcheng5** [comment]
```r
widget %>%
  onRender("function(el, x) { ... }")
```

Not bad at all... I think I'll do that...

jjallaire:
I like hook / addRenderHook as well

jcheng5:
Are we good with the name `onRender`? If so, I think this is ready to merge.

timelyportfolio:
Works for me.  Name is sort-of intuitive and JS-like.  Also, I can't think of any conflicts with other R packages.

jjallaire:
Yes, agreed. Good to merge!

J

PR #173, opened 2015-12-02
Adapt docs for new JS binding style

jcheng5:
Please don't merge until next version of htmlwidgets ships to CRAN.

See also https://github.com/jjallaire/sigma/pull/6


ramnathv:
Thanks @jcheng5. I will update the `scaffoldWidget` function to make use of the new instance-bound definition and send a PR.

jcheng5:
Ohhh... right, forgot about that. Thanks.

As long as you're in the scaffoldWidget code, some of the inlined R code template is confusing roxygen2 5.0.1 (it's fine with the latest 4.x). Maybe move code templates to inst/, with a .txt extension?

ramnathv:
Yes. That is a good idea. I will move the template code over to `inst` and just read it in the `scaffoldWidget` code. Thanks for the pointer.

PR #175, opened 2016-01-07
Move templates to files

ramnathv:
This PR does a couple of things:

1. Moves scaffold code into templates, so that it doesn't trip later versions of `roxygen2`.
2. Adds scaffold template for the new instance based specification of widgets.
3. Bumps version number.

This PR only touches scaffolding code. So, we only need to confirm that `htmlwidgets::scaffoldWidget` works as before. I have tested it locally and everything seems to be working well. But it would be good if a few of you could independently test it and confirm before we merge. @timelyportfolio @jjallaire @jcheng5 @yihui

### DIFF
--- a/R/scaffold.R
+++ b/R/scaffold.R
@@ -31,58 +31,9 @@ scaffoldWidget <- function(name, bowerPkg = NULL, edit = interactive()){
 }
 
 addWidgetConstructor <- function(name, package, edit){
-  tpl <- "#' <Add Title>
-#'
-#' <Add Description>
-#'
-#' @import htmlwidgets
-#'
-#' @export
-%s <- function(message, width = NULL, height = NULL) {
-
-  # forward options using x
-  x = list(
-    message = message
-  )
-
-  # create widget
-  htmlwidgets::createWidget(
-    name = '%s',
-    x,
-    width = width,
-    height = height,
-    package = '%s'
-  )
-}
-
-#' Shiny bindings for %s
-#'
-#' Output and render functions for using %s within Shiny
-#' applications and interactive Rmd documents.
-#'
-#' @param outputId output variable to read from
-#' @param width,height Must be a valid CSS unit (like \\code{'100\\%%'},
-#'   \\code{'400px'}, \\code{'auto'}) or a number, which will be coerced to a
-#'   string and have \\code{'px'} appended.
-#' @param expr An expression that generates a %s
-#' @param env The environment in which to evaluate \\code{expr}.
-#' @param quoted Is \\code{expr} a quoted expression (with \\code{quote()})? This
-#'   is useful if you want to save an expression in a variable.
-#'
-#' @name %s-shiny
-#'
-#' @export
-%sOutput <- function(outputId, width = '100%%', height = '400px'){
-  shinyWidgetOutput(outputId, '%s', width, height, package = '%s')
-}
-
-#' @rdname %s-shiny
-#' @export
-render%s <- function(expr, env = parent.frame(), quoted = FALSE) {
-  if (!quoted) { expr <- substitute(expr) } # force quoted
-  shinyRenderWidget(expr, %sOutput, env, quoted = TRUE)
-}
-"
+  tpl <- paste(readLines(
+    system.file('templates/widget_r.txt', package = 'htmlwidgets')
+  ), collapse = "\n")
 
   capName = function(name){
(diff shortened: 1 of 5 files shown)


inst/templates/widget_js_new.txt
```diff
@@ -0,0 +1,19 @@
+HTMLWidgets.widget({
+
+  name: '%s',
+
+  type: 'output',
+
+  factory: function(el, width, height) {
+
+    return {
```
jjallaire: Might be nice to add some spacing around the function as we had in the previous template

R/scaffold.R
```diff
@@ -128,32 +79,15 @@ addWidgetYAML <- function(name, bowerPkg, edit){
...
-
-  renderValue: function(el, x, instance) {
-
-    el.innerText = x.message;
-
-  },
-
-  resize: function(el, width, height, instance) {
-
+  if (packageVersion('htmlwidgets') <= '0.5.2'){
```
jjallaire: Since this code is running in 0.5.3 would this condition always be FALSE?

inst/templates/widget_js_new.txt
jjallaire: Maybe also some one line comments on what each function is for (we do a lot of this in the rstudio templates and people seem to really appreciate it)

ramnathv: Good point. I will add spacing as well as comments to indicate what each of the functions do.

R/scaffold.R
ramnathv: Yes. This check will be redundant. I am not sure what I was thinking :smiley:

inst/templates/widget_r.txt
```diff
@@ -0,0 +1,52 @@
...
+#' @param expr An expression that generates a %s
+#' @param env The environment in which to evaluate \\code{expr}.
+#' @param quoted Is \\code{expr} a quoted expression (with \\code{quote()})? This
+#'   is useful if you want to save an expression in a variable.
+#'
+#' @name %s-shiny
+#'
+#' @export
+%sOutput <- function(outputId, width = '100%%', height = '400px'){
+  shinyWidgetOutput(outputId, '%s', width, height, package = '%s')
```
timelyportfolio: since we don't import `htmlwidgets` with these Shiny functions, should this not be

```
htmlwidgets::shinyWidgetOutput
```

?

inst/templates/widget_r.txt
```diff
@@ -0,0 +1,52 @@
...
+}
+
+#' @rdname %s-shiny
+#' @export
+render%s <- function(expr, env = parent.frame(), quoted = FALSE) {
+  if (!quoted) { expr <- substitute(expr) } # force quoted
+  shinyRenderWidget(expr, %sOutput, env, quoted = TRUE)
```
timelyportfolio: also need `htmlwidgets::`

```
htmlwidgets::shinyRenderWidget(
```

inst/templates/widget_r.txt
```diff
@@ -0,0 +1,52 @@
...
+  )
+}
+
+#' Shiny bindings for %s
+#'
+#' Output and render functions for using %s within Shiny
+#' applications and interactive Rmd documents.
+#'
+#' @param outputId output variable to read from
+#' @param width,height Must be a valid CSS unit (like \\code{'100\\%%'},
```
timelyportfolio: when I used I got an error for mismatched braces; changing to `\%` seemed to fix, but I'm not sure if this is just an issue with Windows or my local devel environment

inst/templates/widget_r.txt
ramnathv: Yes. I believe the namespacing will be required. I will add that. Thanks for catching it.

inst/templates/widget_r.txt
ramnathv: Can you use `readLines` to directly read the template and see if you are getting the error? It works for me, but I am on Mac OS.

ramnathv: Oh you know what, I think I needed the extra `\` since I had the template inline earlier. Now that I have moved it into a file, I can just stick with a single `\` probably.

jjallaire:
It's not pushed yet because I'm on an airplane that won't let me talk to GitHub, but I've incorporated this pull request (and the various tweaks/comments) into another commit so we're now using template based scaffolding. I'll close this PR, you should see the commits in a few hours!

PR #177, opened 2016-01-11
Add an inline argument for shiny output widgets

cpsievert:
This implementation is modeled after `shiny::imageOutput()`

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -180,7 +180,7 @@ toHTML <- function(x, standalone = FALSE, knitrOptions = NULL) {
 }
 
 
-widget_html <- function(name, package, id, style, class, ...){
+widget_html <- function(name, package, id, style, class, inline = FALSE, ...){
 
   # attempt to lookup custom html function for widget
   fn <- tryCatch(get(paste0(name, "_html"),
@@ -191,6 +191,8 @@ widget_html <- function(name, package, id, style, class, ...){
   # call the custom function if we have one, otherwise create a div
   if (is.function(fn)) {
     fn(id = id, style = style, class = class, ...)
+  } else if (inline) {
+    tags$span(id = id, style = style, class = class)
   } else {
     tags$div(id = id, style = style, class = class)
   }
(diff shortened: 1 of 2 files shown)


cpsievert:
To be clear, the purpose of this PR is to make it easier to put inline widgets in shiny apps, like this -- 

http://104.131.111.111:3838/plotlyLinkedBrush/
https://github.com/cpsievert/shiny_apps/blob/master/plotlyLinkedBrush/app.R

ramnathv:
@cpsievert sounds like an easy change to make. On the demo plots that you linked to earlier, where do the inline widgets surface? Will `display:inline-block` on a  `div` not give you what you seek?

cpsievert:
Yea, I suppose you could do the same with something like:

```r
htmltools::div(style = "display:inline-block", plotlyOutput("x", width = 400, height = 250))
```

 but the `inline` argument would be a convenient for users with limited HTML/CSS knowledge

R/htmlwidgets.R
```diff
@@ -329,16 +333,18 @@ createWidget <- function(name,
 #' @name htmlwidgets-shiny
 #'
 #' @export
-shinyWidgetOutput <- function(outputId, name, width, height, package = name) {
+shinyWidgetOutput <- function(outputId, name, width, height, inline = FALSE,
+                              package = name) {
```
jcheng5: Maybe the new arg should be last, for backward compatibility? Or do we think literally nobody is specifying the package?

timelyportfolio: I think most probably go with the template from `scaffoldWidget` which currently looks and is proposed to look like [this](https://github.com/ramnathv/htmlwidgets/blob/move-templates-to-files/inst/templates/widget_r.txt#L42):

```
#' @export
%sOutput <- function(outputId, width = '100%%', height = '400px'){
  shinyWidgetOutput(outputId, '%s', width, height, package = '%s')
}
```

jcheng5: Ack. So it turns out that my own Leaflet package is passing the package argument explicitly, without `package =`, and so breaks with the current htmlwidgets master. @cpsievert, any objection to my putting `inline = FALSE` last?

jcheng5: Offending line here: https://github.com/rstudio/leaflet/blob/f89f66d19149ead272952de8de808f7127fe4a14/R/shiny.R#L23

cpsievert: > any objection to my putting `inline = FALSE` last?

Fine by me @jcheng5! Would you like another pull request

jcheng5: Nah I'll take care of it. I have it right in front of me. Thanks!

PR #183, opened 2016-02-07
Avoid copying the js binding to a temp dir

yihui:
Because otherwise it is impossible to cache code chunks that involve HTML widgets: the temp dir path changes between R sessions and gets deleted every time R quits, so the dependency object will not be reusable for the next R session

Note this PR relies on https://github.com/rstudio/htmltools/pull/48

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,7 +17,7 @@ Description: A framework for creating HTML widgets that render in various
 License: MIT + file LICENSE
 VignetteBuilder: knitr
 Imports:
-    htmltools (>= 0.3.3),
+    htmltools (>= 0.3),
     jsonlite (>= 0.9.16),
     yaml
 Suggests:
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,7 +17,7 @@ Description: A framework for creating HTML widgets that render in various
 License: MIT + file LICENSE
 VignetteBuilder: knitr
 Imports:
-    htmltools (>= 0.2.6),
+    htmltools (>= 0.3.2),
     jsonlite (>= 0.9.16),
     yaml
 Suggests:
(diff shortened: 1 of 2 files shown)


jcheng5:
Shouldn't the decision of whether the binding die should be copied, be a property decided by the caller of htmlDependency and not by the environment? I'd be more OK with hard coding it to FALSE than controlling it with an option.

jcheng5:
Oh I see, you're just porting existing logic. Let's get rid of that option unless anyone else has an objection.

yihui:
I completely agree with you. I do not see the point of copying other files. Unless @ramnathv has anything else on mind (#136), I'll just hard-code it to `all_files = FALSE`.

Are you okay with the htmltools PR?

jjallaire:
If this PR takes our htmltools dependency up to v.next of htmltools (and
this requires a CRAN release of htmltools) let's wait on it. I am hoping to
submit a new version of htmlwidgets to CRAN soon and don't want it blocked
by htmltools (or the building of binaries on OSX for htmltools, which can
take a long time. @yihui, if it's okay with you we'll do another
htmlwidgets release in 30-60 days where we can pick up this and other
pending changes.

ramnathv:
@yihui I want to test this one out thoroughly, to make sure that it does not break anything for me. I will revert back in a couple of days.

yihui:
@jjallaire I just removed the htmltools version 0.3.3 dependency, and htmlwidgets will still use the old trick when htmltools version is smaller than 0.3.3. So this should be good to go even without a new release of htmltools on CRAN.

jjallaire:
Okay, sounds good!

jjallaire:
@ramnathv are you okay if I merge this? (preparing for a CRAN release)

ramnathv:
Yes @jjallaire. @yihui made a change that retains the older behavior, which I believe I am the only one using. So this one is good to be merged.

PR #189, opened 2016-02-19
Fix #151: In JS bindings, resize() can't call renderValue()

### LINKED ISSUE #151: In JS bindings, resize() can't call renderValue()

yihui:
Rebased #151 against master so Joe's two commits can be merged. I have tested it with https://github.com/rstudio/DT/issues/238 and the bug was indeed fixed.

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -437,18 +437,27 @@
       // we're not delegating, then that call will go to the Shiny
       // version instead of the htmlwidgets version.
 
-      // Merge defaults into the definition; don't mutate the original definition.
-      // The base object is a Shiny output binding if we're running in Shiny mode,
-      // or an empty object if we're not.
-      var bindingDef = extend(defaults, definition);
+      // Merge defaults with definition, without mutating either.
+      var bindingDef = extend({}, defaults, definition);
+
+      // This object will be our actual Shiny binding.
       var shinyBinding = new Shiny.OutputBinding();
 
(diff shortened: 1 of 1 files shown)


jjallaire:
@jcheng5 I don't think any of us know this code as well as you do so we will leave it to you to merge when you feel confident the change is good. Do you feel any particular risk in our making this change then submitting to CRAN soon after?

jjallaire:
@jcheng5 A reminder to merge this today (or declare that we shouldn't merge because it's too risky right before a CRAN release).

PR #190, opened 2016-02-23
Move new `inline` param, for back-compat

jcheng5:
The introduction of the new `inline` parameter broke any
code that assumed the 5th argument was `package`. This
isn't the case for scaffold-generated widgets, but at
least the leaflet package made this assumption.


jcheng5:
Repro:

```r
> leaflet::leafletOutput("map")
Error in if (inline) "display: inline-block;" else "" : 
  argument is not interpretable as logical
```

With this fix:

```r
> leaflet::leafletOutput("map")
[1] "<div id=\"map\" style=\"width:100%; height:400px; \" class=\"leaflet html-widget html-widget-output\"></div>"
```

timelyportfolio:
thanks @jcheng5

PR #191, opened 2016-02-25
Pass knitr options to saveWidget()

yihui:
I cannot say for sure for now, but it is very likely that I will need this for the screenshotting work that I have been doing.


jjallaire:
Going to hold this PR for until after we get to v0.6 CRAN (my guess is that there will be some iteration and perhaps the addition of a saveWidgetAsImage function to go along with this).

yihui:
Do you think this PR is ready to go now? I can provide a new `saveWidgetAsImage()` function based on webshot (which is on its way to CRAN now) in a separate PR if you want.

PR #195, opened 2016-03-22
Support shown/hidden.bs.collapse events for Bootstrap collapse

yihui:
Fixes rstudio/DT#204

### DIFF
--- a/inst/NEWS
+++ b/inst/NEWS
@@ -3,6 +3,9 @@ htmlwidgets 0.7 (unreleased)
 
 * Pass knitr options to saveWidget
 
+* The resize handler also works for the JavaScript events `shown.bs.collapse`
+  and `hidden.bs.collapse` now so that widgets inside the Bootstrap collapse
+  class can be displayed
 
 htmlwidgets 0.6
 -----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ramnathv:
This looks good to me. However, instead of increasing the list of events that trigger the `resizeHandler`, cant we add a HTMLWidgets method that allows one to register custom events that trigger the resize handler? This could be a property on the widget that widget authors can customize? Or maybe it could be specific to a widget instance, so that a user can register custom events. @jcheng5 @jjallaire @timelyportfolio  what do you think?

jjallaire:
I agree that it would be nice for widgets to be able to provide a list of jquery event selectors which result in a resize, however given the number of widgets (now 30 on CRAN alone!) and the fact that the cases where a resize is required are to a large extent enumerable I think it's better if we can handle this centrally so widgets don't have to discover these cases from first principles.

ramnathv:
Sounds good @jjallaire and @yihui

timelyportfolio:
I have been thinking a lot about this lately, and I hate having to inject framework-specific event handlers in `htmlwidgets` code, but I can't think of any way around it.  Perhaps, we could work on separating these out some way, so they are clear.

PR #199, opened 2016-04-09
call widget.resize when slides are shown

jjallaire:
This PR hoists a workaround that I have in dygraphs into the core htmlwidgets. Here is the original code in dygraphs:

https://github.com/rstudio/dygraphs/blob/master/inst/htmlwidgets/dygraphs.js#L131-L144

The issue is that for dygraphs when they start out as display:none (which is the case in ioslides and revealjs among perhaps others) their width and height is 0. This means that when the slide they are hosted on is shown they appear invisible.

Most widgets don't have this problem, but at least one developed by @bwlewis does and it must also also affect others. @bwlewis Could you confirm that this resolves the problems you have been seeing?

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -588,20 +588,13 @@
               "hidden.htmlwidgets hidden.bs.tab.htmlwidgets hidden.bs.collapse.htmlwidgets",
               resizeHandler
             );
-          }
 
-          // This is needed for the specific case of ioslides, which
-          // flips slides between display:none and display:block.
-          // Ideally we would not have to have ioslide-specific code
-          // here, but rather have ioslides raise a generic event,
-          // but the rmarkdown package just went to CRAN so the
-          // window to getting that fixed may be long.
-          if (window.addEventListener) {
-            // It's OK to limit this to window.addEventListener
-            // browsers because ioslides itself only supports
-            // such browsers.
-            on(document, "slideenter", resizeHandler);
-            on(document, "slideleave", resizeHandler);
+            // subscribe to custom shown event fired by ioslides and reveal.js (and
+            // perhaps other slide frameworks). This is necessary because some widgets
+            // (e.g. dygraphs) which start out as display:none have height == 0 and
+            // width == 0 and this doesn't change when it becomes visible
+            window.jQuery(el).closest('slide').on('shown', resizeHandler);
+            window.jQuery(el).closest('section.slide').on('shown', resizeHandler);
           }
         }
 


timelyportfolio:
This is great, but what if we don't have `jQuery`?

jjallaire:
It doesn't need to be written as a jquery plugin that's just how I did it
for dygraphs.

jjallaire:
Oh sorry you are talking about the resize thing (not the other issue I
posted). I don't know how to do this quite as cleanly without jquery
(although I'm sure it's possible). The previous workaround called resize on
every widget in the entire slideshow every time a slide was shown/hidden.
The jquery allowed me to easily attach the event handler to *just* the
slide containing the widget.

Net: at a practical level jquery will be present 98% of the time so this
workaround will take effect in those cases. If someone else wants to write
the equivalent non-jquery code I'm all for it, but if no one has the time
or motivation to do that right now I don't think it should prevent us from
taking this fix in the meantime.

timelyportfolio:
On my phone but this http://clubmate.fi/jquerys-closest-function-and-pure-javascript-alternatives/ appears to offer a reasonable alternative to closest.  I will volunteer to try this or some other ideas I have so that this will be available without jQuery.

ramnathv:
This looks good to me. And I agree with @jjallaire that we should roll with the `jquery` solution for now and if we end up creating a vanilla js equivalent, we can swap it out for this code. Do we need to test this with a shiny runtime @jcheng5 ?

jcheng5:
Hmmm. I don't understand why this code changes anything--the registration right above (to `"shown.htmlwidgets"`) should cover all `"shown"` events in the document (unless someone is calling `event.stopPropagation()` on the `shown` event, which seems unlikely?).

@jjallaire Do you have a repro case I can look at?

jjallaire:
Indeed I can't repro any longer as it looks to me like the global shown
event is taking care of it. Here's the change where this was fixed (right
before the CRAN release of last May):

https://github.com/ramnathv/htmlwidgets/commit/b772404787e12570434517da0d0415d457e48e4a

I put the dygraphs code in long before that so for that case this fix is
essentially already in.

However, I was seeing the same problem with revealjs even with that commit.
This change went in after the May CRAN release (and I would have been
testing revealjs without it) so perhaps the revealjs issue was due to this
bug not yet being fixed:

https://github.com/ramnathv/htmlwidgets/commit/1582c461667b328a2d8b786afb77a12191ed50a4

So it looks like none of this is necessary *except* that @bwewis still has
a repro of this in ioslides with one of his widgets. @bwlewis could you
post a simple version of this repro?

bwlewis:
Will do, I just got back. I will first make sure I was using latest versions of everything, perhaps that machine had something old by chance. More soon.

PR #202, opened 2016-04-16
Add ability for onRender to take arbitrary R data

jcheng5:
This makes it much easier for `onRender` custom JS functions to operate on data that exists in R. Rather than having to do something like:

```r
myWidget() %>%
  onRender(sprintf("function(el, x) { var data = %s; this.addRecords(data); }", toJSON(iris)))
```

you can do:

```r
myWidget() %>%
  onRender("function(el, x, data) { this.addRecords(data); }", data = iris)
```

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -76,15 +76,23 @@ appendContent <- function(x, ...) {
 #'
 #' @param x An HTML Widget object
 #' @param jsCode Character vector containing JavaScript code (see Details)
+#' @param data An additional argument to pass to the \code{jsCode} function.
+#'   This can be any R object that can be serialized to JSON. If you have
+#'   multiple objects to pass to the function, use a named list.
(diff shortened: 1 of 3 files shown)


jcheng5:
P.S. I tried changing the example from `\dontrun` to `\donttest`, but R CMD check then complains that the leaflet dependency isn't declared. Doesn't seem like a good enough reason to take leaflet as a Suggests. If anyone knows a way around this I'd love to hear it--it's a nice pair of examples and it's a shame not to be able to do `example(onRender)`.

jcheng5:
See the heatmap example here:
https://gist.github.com/jcheng5/c084a59717f18e947a17955007dc5f92

image

jjallaire:
/This looks good to me. @ramnathv ?

ramnathv:
This looks good @jcheng5. We can merge these changes in 👍

PR #213, opened 2016-07-12
Use uuid package rather than stats::runif for generating widget IDs

jjallaire:


jjallaire:
One downside to this approach is that every time html is rendered distinct html id's will be generated. There's no real downside to this other than annoyance when reviewing VCS diffs. The alternative would be to just use a larger `runif`value to reduce the chance of collisions.

@jcheng5 Let us know what you think the right approach is here.

ramnathv:
Another approach is to generate a hash based on `x`. This will add an extra computation to the mix, but will make things VCS friendly. Another idea is to turn this into an option that can be set by the user rendering the widgets. We can make three options available (current approach, runif, and digest), plus user defined function. This might be overkill, but wanted to lay out the possibility.

jjallaire:
Would the contents of x be available at the time the ID is getting
generated? (I think the ID is generated before x is available)

ramnathv:
Based on this https://github.com/ramnathv/htmlwidgets/pull/213/commits/2ede11db1ebfa53cb573c3db2242020d86040a67, it seems like `x` is available, since we use the random id only when an `elementId` is not specified.

jjallaire:
Okay, I wonder though if "x" is guaranteed to be unique (the same widget
instantiated with identical arguments would give the same digest).

Let's wait to hear what @jcheng5 thinks about taking the runif number up as
an alternative to uuids.

jjallaire:
That's correct, the lack of stability is indeed the same, but in some cases this is mediated by calling set.seed (e.g. I do this for the htmlwidgets website). 

I think moving away from set.seed / runif seems sensible, although I'm not sure about using the mechanism you do in Shiny in preference to uuids. From what I can see I'd have to copy and paste at least what's below (perhaps there's a couple more bits I'm not seeing). Creating a unique ID with 16 bytes would then yield me an ID very close in length to a uuid. I'm not sure what the advantage would be? (the uuid package has no dependencies).

If you feel strongly that there are big disadvantages to uuids maybe we can wrap what you are doing in Shiny into it's own package rather than copying and pasting the code. Let me know what you think.

```r
createUniqueId <- function(bytes, prefix = "", suffix = "") {
  withPrivateSeed({
    paste(
      prefix,
      paste(
        format(as.hexmode(sample(256, bytes, replace = TRUE)-1), width=2),
        collapse = ""),
      suffix,
      sep = ""
    )
  })
}

# Temporarily set x in env to value, evaluate expr, and
# then restore x to its original state
withTemporary <- function(env, x, value, expr, unset = FALSE) {

  if (exists(x, envir = env, inherits = FALSE)) {
    oldValue <- get(x, envir = env, inherits = FALSE)
    on.exit(
      assign(x, oldValue, envir = env, inherits = FALSE),
      add = TRUE)
  } else {
    on.exit(
      rm(list = x, envir = env, inherits = FALSE),
      add = TRUE
    )
  }

  if (!missing(value) && !isTRUE(unset))
    assign(x, value, envir = env, inherits = FALSE)
  else {
    if (exists(x, envir = env, inherits = FALSE))
      rm(list = x, envir = env, inherits = FALSE)
  }
  force(expr)
}

.globals$ownSeed <- NULL
# Evaluate an expression using Shiny's own private stream of
# randomness (not affected by set.seed).
withPrivateSeed <- function(expr) {
  withTemporary(.GlobalEnv, ".Random.seed",
    .globals$ownSeed, unset=is.null(.globals$ownSeed), {
      tryCatch({
        expr
      }, finally = {
        .globals$ownSeed <- getExists('.Random.seed', 'numeric', globalenv())
      })
    }
  )
}
```

jjallaire:
I guess one advantage might be that we might convince ourselves that we can use e.g. 8 bytes instead of 16 which would yield us an ID roughly half the size of a uuid. But the size might not matter considering that it's never user visible (I guess debugging with shorter ids might be one consideration, but it seems like IDs are usually copied and pasted).

jjallaire:
If we don't like the long uuids we could also just shorten them (in RStudio we hash them in order to shorten them): https://github.com/rstudio/rstudio/blob/ebb054efe3e650de818a251b11087311e5169ed8/src/cpp/core/system/System.cpp#L86-L90

We already recursively depend on digest (via htmltools) so it wouldn't be an additional dependency.

Of course I'm far from an expert on random identifiers / crypto / hashing / etc. so I might be thinking about this all wrong.

jcheng5:
Yes, the advantage of createRandomId is just that we can use whatever number of bytes we feel is appropriate. There's a birthday paradox calculator here: https://lazycackle.com/Probability_of_repeated_event_online_calculator__birthday_problem_.html where `m` is some number of widgets, `n` is the number of possible IDs (so 10000 today, 2^128 with a uuid, and 2^(bytes*8) for createRandomId).

This shows that n=10000 is indeed too low, with 100 widgets the chance of collision is 39%. But even with just 4 bytes of entropy (8 character long id) the odds drop to 0.0001% for 100 widgets, or 0.01% for 1000 widgets.

So I think it's potentially worth dropping the number of bytes from e.g. `"49f4ca7b007b00100000064284d6e57b"` to `"f27b003a"`, whether we do that by shortening the uuid or using createRandomId, which really just boils down to this:

```r
    paste(
      format(as.hexmode(sample(256, bytes, replace = TRUE)-1), width=2),
      collapse = "")
```

(I'll address reproducibility in a separate comment)

jcheng5:
I think it'd be nice to have reproducible id names--I personally often take the time to undo spurious element ID name changes, and while we can't fix all cases (i.e. inserting a new widget in the middle of a page) a set.seed-like thing can certainly help.

The impl in Shiny is a little verbose (I had written it to support not only random seeds, but any other kind of value we might want to temporarily override), but it comes down to this (where `.private` is just an env for storing global variables for the package):

```r
.private <- new.env(parent = emptyenv())

createUniqueId <- function(bytes) {
  # Note what the system's random seed is before we start, so we can restore it after
  sysSeed <- .GlobalEnv$.Random.seed
  # Replace system seed with our own seed
  if (!is.null(.private$ownSeed)) {
    .GlobalEnv$.Random.seed <- .private$ownSeed
  }
  on.exit({
    # Change our own seed to match the current seed
    .private$ownSeed <- .GlobalEnv$.Random.seed
    # Restore the system seed--we were never here
    .GlobalEnv$.Random.seed <- sysSeed
  })

  paste(
    format(as.hexmode(sample(256, bytes, replace = TRUE)-1), width=2),
    collapse = "")
}
```

So just swapping a couple of variables before we start, and swapping them back when we end.

The you do something very similar to allow the user to call the equivalent of `set.seed` on our own seed:

```r
setOwnSeed <- function(seed, kind = NULL, normal.kind = NULL) {
  sysSeed <- .GlobalEnv$.Random.seed
  on.exit({
    .private$ownSeed <- .GlobalEnv$.Random.seed
    .GlobalEnv$.Random.seed <- sysSeed
  })
  set.seed(seed, kind = kind, normal.kind = normal.kind)
}
```

jjallaire:
Okay, excellent! I just revised the implementation to use your function. Let me know if this looks okay.

jcheng5:
LGTM!

PR #214, opened 2016-07-12
Ensure that scaffoldWidget opens files correctly within RStudio

jjallaire:
RStudio overrides the `utils::file.edit` function but only when invoked unqualified within the global environment. This PR invokes file.edit in a way that will result in correct binding to the rstudio implementation when running within RStudio.

We could improve this by importing the rstudioapi package and calling the RStudio API directly, however I that's a bit more heavyweight and I don't think warranted. Note that the call to `file.edit` will fail if `utils` isn't attached to the search path but I think this will be exceptionally rare.

### DIFF
--- a/R/scaffold.R
+++ b/R/scaffold.R
@@ -48,7 +48,7 @@ addWidgetConstructor <- function(name, package, edit){
   } else {
     message(file_, " already exists")
   }
-  if (edit) file.edit(file_)
+  if (edit) fileEdit(file_)
 }
 
 addWidgetYAML <- function(name, bowerPkg, edit){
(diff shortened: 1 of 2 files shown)


ramnathv:
Thanks @jjallaire. This will not have any impact outside of RStudio right? Any ideas why this worked before, but fails now?

timelyportfolio:
Thanks @jjallaire.  Looks good to me. Should we increment version?

jjallaire:
No, it won't have any impact outside of RStudio.

To be honest I have no idea why it worked before and fails now (perhaps a
change in the way R handles the "editor" option?).

jjallaire:
Let's increment the version after merging.

daattali:
thanks @jjallaire

PR #217, opened 2016-07-14
pass along "style" param from htmlwidgets html to custom html

daattali:
I think the "style" param is necessary in order to have non-default width/height eg. `fooWidget(width = "50%")` without passing along the style, the width won't actually be 50%.

And I'm wondering if we need to use the `...` arguments at all?  Or is it always safe to ignore them?

### DIFF
--- a/vignettes/develop_advanced.Rmd
+++ b/vignettes/develop_advanced.Rmd
@@ -178,7 +178,7 @@ Typically the HTML "housing" for a widget is just a `<div>` element, and this is
 
 ```r
 sparkline_html <- function(id, style, class, ...){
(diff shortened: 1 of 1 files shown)


jcheng5:
Maybe a bad example, span tags don't respect width anyway... (unless you change their display mode to e.g. inline-block)

daattali:
@jcheng5 true, I neglected to realize this is a span specifically. Maybe the example should be left as-is but there should be a note nearby that says that the `style` param is also suggested to be passed through if they want the user-supplied height+width to go through.   Or maybe I'm just wrong altogether - I only started looking at htmlwidgets this week and am a complete noob...

Issue #234, opened 2016-09-23
'createWidgetId' sets a NULL .Random.seed if none already set


kevinushey:
Repro:

```R
htmlwidgets:::createWidgetId()
print(.Random.seed)
rnorm(10)
```

gives

```
> htmlwidgets:::createWidgetId()
[1] "ed93cb48d364b49da97c"
print(.Random.seed)
rnorm(10)
> print(.Random.seed)
NULL
rnorm(10)
> rnorm(10)
Warning in rnorm(10) :
  '.Random.seed' is not an integer vector but of type 'NULL', so ignored
 [1] -0.2118466 -1.6132225 -1.6132225 -1.6132225 -1.7495102 -6.4517136
 [7] -5.7383473 -1.6132225 -1.6132225 -1.6132225
```

Note the scary-looking warning printed in response to the `rnorm(10)` call. `createWidgetId` should only reset the `.Random.seed` if it existed and was an integer vector.

PR #235, opened 2016-09-23
only restore .Random.seed when non-NULL

kevinushey:
Closes https://github.com/ramnathv/htmlwidgets/issues/234.


jjallaire:
LGTM. @jcheng5 Could you take a look at this as well?

PR #237, opened 2016-09-26
Fix #236: onRender doesn't work with Shiny

### LINKED ISSUE #236: onRender example doesn't work in a shiny app
I've never used onRender() and I've never used leaflet until today, so it's very possible that I'm doing something wrong.  But it seems to me that since the `onRender()` example shows how to use it on a leaflet map, and leaflet maps can be inside shiny apps, then the same example should also work within a shiny app. Perhaps I'm wrong though?

This code is taken straight out of the documentation for `onRender` and it works:

```
meh <- "&#x1F610;";
yikes <- "&#x1F628;";

df <- data.frame(
  lng = quakes$long,
  lat = quakes$lat,
  html = ifelse(quakes$mag < 5.5, meh, yikes),
  stringsAsFactors = FALSE
)

leaflet() %>% addTiles() %>%
  fitBounds(min(df$lng), min(df$lat), max(df$lng), max(df$lat)) %>%
  onRender("
    function(el, x, data) {
      for (var i = 0; i < data.lng.length; i++) {
        var icon = L.divIcon({className: '', html: data.html[i]});
        L.marker([data.lat[i], data.lng[i]], {icon: icon}).addTo(this);
      }
    }
  ", data = df)
```

However, if I place it inside a shiny app, it doesn't work:

```
library(shiny)
library(leaflet)
library(htmlwidgets)

ui <- fluidPage(
  leafletOutput("mymap")
)

  output$mymap <- renderLeaflet({
    meh <- "&#x1F610;";
    yikes <- "&#x1F628;";

    df <- data.frame(
      lng = quakes$long,
      lat = quakes$lat,
      html = ifelse(quakes$mag < 5.5, meh, yikes),
      stringsAsFactors = FALSE
    )

    leaflet() %>% addTiles() %>%
      fitBounds(min(df$lng), min(df$lat), max(df$lng), max(df$lat)) %>%
      onRender("
               function(el, x, data) {
               for (var i = 0; i < data.lng.length; i++) {
               var icon = L.divIcon({className: '', html: data.html[i]});
               L.marker([data.lat[i], data.lng[i]], {icon: icon}).addTo(this);
               }
               }
               ", data = df)
  })
}

shinyApp(ui, server)
```

Specifically, I see the following error in the javascript console:

```
leaflet.js:1516 Uncaught TypeError: Cannot read property 'getGroupNameFromLayerGroup' of undefined
```

The exact javascript code in that line is

```
map.layerManager.getGroupNameFromLayerGroup(e.layer)
```

So I then placed a simple `console.log(this.layerManager)` and I saw that indeed when running the map in the console it prints out a LayerManager object, but inside a shiny app it's undefined.

I'm not sure if this is a bug or expected, but to someone with no prior experience like myself it seems like unexpected behaviour

jcheng5:
The `onRender` hook was being called after initialize, not after render. My bad!

cc @daattali @bhaskarvk

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: htmlwidgets
 Type: Package
 Title: HTML Widgets for R
-Version: 0.7
+Version: 0.8
 Date: 2016-08-01
 Authors@R: c(
     person("Ramnath", "Vaidyanathan", role = c("aut", "cph")),
--- a/inst/NEWS
+++ b/inst/NEWS
@@ -1,3 +1,9 @@
+htmlwidgets 0.8 (unreleased)
+-----------------------------------------------------------------------
+
+* `onRender` hooks were firing too early when used in Shiny apps.
+
+
 htmlwidgets 0.7
 -----------------------------------------------------------------------
 
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -511,10 +511,10 @@
               el.offsetHeight);
             elementData(el, "init_result", result);
           }
-          evalAndRun(data.jsHooks.render, elementData(el, "init_result"), [el, data.x]);
         }
         Shiny.renderDependencies(data.deps);
         bindingDef.renderValue(el, data.x, elementData(el, "init_result"));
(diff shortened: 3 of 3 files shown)


jcheng5:
@ramnathv @jjallaire This is ready for merging.

Issue #239, opened 2016-10-30
Problem saving widget with custom html


johndharrison:
I am trying to add custom html to an htmlwidget as per
https://cran.r-project.org/web/packages/htmlwidgets/vignettes/develop_advanced.html#custom-widget-html

To replicate the issue I have create a simple htmlwidget package with minimal code using 
```
htmlwidgets::scaffoldWidget("mywidget")
```
I added some custom html for the widget https://github.com/johndharrison/simplewidget/commit/23b38b3c392d959aabe37d4c7b10f09df25af106

The widget displays correctly and the custom html functions as expected. However `htmlwidgets::saveWidget` does not give the expected output:
```
install.github("johndharrison/simplewidget")
library(simplewidget)
library(xml2)
library(magrittr)
sw <- simplewidget("hello, world")
tFile <- tempfile(fileext = ".html")
htmlwidgets::saveWidget(sw, file = tFile)
browseURL(tFile)
xml2::read_html(tFile) %>% 
  xml2::xml_find_all("//div[@id='htmlwidget_container']") %>% 
  as.character
[1] "<div id=\"htmlwidget_container\">\n<div>\n<pre><code>&lt;h1 id=\"someid\"&gt;My header&lt;/h1&gt;\n&lt;div id=\"htmlwidget-00d700b400000a3685d7\" style=\"width:960px;height:500px;\" class=\"simplewidget html-widget\"&gt;&lt;/div&gt;</code></pre>\n</div>\n</div>"
```

johndharrison:
The issue seems to be with pandoc and indentation introduced by `htmltools::tags`. Using `htmltools::renderTags` to remove the indentation https://github.com/johndharrison/simplewidget/commit/3ecd3bcfa73a4711b04a799dab8b37a9f9e12ffc seems to solve the issue.

PR #240, opened 2016-10-30
Render user custon html for widget.

johndharrison:
Address #239 issue with custom html via "_html" not being rendered.

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -226,7 +226,9 @@ widget_html <- function(name, package, id, style, class, inline = FALSE, ...){
 
   # call the custom function if we have one, otherwise create a div
   if (is.function(fn)) {
-    fn(id = id, style = style, class = class, ...)
+    renderTags(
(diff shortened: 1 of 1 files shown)


R/htmlwidgets.R
```diff
@@ -226,7 +226,9 @@ widget_html <- function(name, package, id, style, class, inline = FALSE, ...){
 
   # call the custom function if we have one, otherwise create a div
   if (is.function(fn)) {
-    fn(id = id, style = style, class = class, ...)
+    renderTags(
```
jjallaire: I don't know if this is the right fix. It might rather be to surround the results of the function in an `html-preserve` comment (which will also prevent pandoc from touching it). @jcheng5 What's the best thing here?

johndharrison:
Hi @jjallaire sure this was just a temporary fix used on a widget I had. I thought it maybe helpful to provide the pull request. Looking at ?htmltools::htmlPreserve that looks like the tool for the job.

Issue #226, opened 2016-08-19
Combine multiple widgets in a single htmlwidget object.


cuche27:
It would be nice to be able to combine multiple widgets in the single like one can combine multiple static plots in a single one using `par(mfrow = c(...))`, for instance to compare the same graphic generated on different datasets or to visualize different informations about the same data.

For now, to do so one has to create a complete shiny app/gadget or to create a rmarkdown document. But during analysis work, it would be be nice to be able to quickly combine widgets with a simple function. 

I implemented a function that do this using `htmltools` and `miniUI` but it is very hacky and has problems of sizing (0px height on rmarkdown documents). Moreover, the function needs to perform a special processing for each kind of htmlwidget.

```r
#' Combine several interactive plots
#'
#' This function combines different interactive plots in a unique view. It is
#' especially useful in the function \code{\link{manipulateWidget}} to have in
#' the same window several related plots that respond to the same set of
#' controls.
#'
#' @param ...
#'   Elements to combine. They should be htmlwidgets, but they can also be
#'   shiny tags or html object or text
#' @param nrow
#'   Number of rows of the layout.
#' @param ncol
#'   Number of columns of the layout.
#' @param title
#'   Title of the view
#' @param hflex
#'   This argument controls the relative size of each column. For instance, if
#'   the layout has two columns and \code{hflex = c(2,1)}, then the width of the
#'   first column will be twice the one of the second one. If a value is equal
#'   to NA, then the corresponding column will have its 'natural' width, and the
#'   remaining space will be shared between the other columns.
#' @param vflex
#'   Same as hflex but for the height of the rows of the layout.
#'
#' @return
#' Object of class 'combinedHtmlwidgets' which is an extension of 'shiny.tags'.
#'
#' @details
#' The function only allows table like layout : each row has the same number of
#' columns and reciprocally. But it is possible to create more complex layout
#' by nesting combined htmlwidgets. (see examples)
#'
#' @examples
#' if require(plotly) {
#'   data(iris)
#'
#'  combineWidgets(title = "The Iris dataset",
#'    plot_ly(iris, x = Sepal.Length, type = "histogram", nbinsx = 20),
#'    plot_ly(iris, x = Sepal.Width, type = "histogram", nbinsx = 20),
#'    plot_ly(iris, x = Petal.Length, type = "histogram", nbinsx = 20),
#'    plot_ly(iris, x = Petal.Width, type = "histogram", nbinsx = 20)
#'  )
#'
[...]

PR #245, opened 2016-11-18
Easily combine multiple htmlwidgets into a single one

FrancoisGuillem:
Hello,

This pull request resolves issue #226 .

It provides a new function `combineWidgets` that permits to easily combine different htmlwidgets in a single for display in the viewer or for inclusion in a Rmarkdown file. Just pass to it several htmlwidgets and eventually a number of rows and columns, a title, etc. and that's all !

 Here is an example of result:

combinewidget

And the code to produce it:

```r
if (require(plotly)) {
  data(iris)

  # Create a complex layout by nesting combinedWidgets
  combineWidgets(title = "The iris data set: sepals", ncol = 2, colsize = c(2,1),
    plot_ly(iris, x = ~Sepal.Length, y = ~Sepal.Width, type = "scatter",
            mode = "markers", color = ~Species),
    combineWidgets(
      plot_ly(iris, x = ~Sepal.Length, type = "histogram", nbinsx = 20),
      plot_ly(iris, x = ~Sepal.Width, type = "histogram", nbinsx = 20)
    )
  )
}
```

For now, this function is available in a ad-hoc package ( https://github.com/cuche27/combineWidgets ) but I think it would be relevant to include it directly in the `htmlwidgets` package.

Cheers


jjallaire:
This is terrific! Thank you :-)

@jcheng5,  @ramnathv, and @timelyportfolio Could you also take a look to make sure there aren't subtleties that might be missing here?

If you are doing active development on the external package let's wait until that settles down before merging this so you don't have to make changes in two places.

cuche27:
Thanks for the answer. 
For now I have not planned to do more development.

Anyway, if you plan to merge my work,  I prefer to contribute directly to htmlwidgets rather than to develop my own package.

jjallaire:
Okay, that sounds good. Could everyone else take a look at this to make sure there aren't other things we need changed before merging. @cuche27 could you also update NEWS with an item about this feature and DESCRIPTION to add yourself as a contributor and copyright holder?

R/combineWidgets.R
```diff
@@ -0,0 +1,186 @@
...
+  # Relative size of rows and cols
+  rowsize <- rep(rowsize, length.out=nrow)
+  colsize <- rep(colsize, length.out = ncol)
+
+  # Get the html ID of each widget
+  elementId <- sapply(widgets[1:ncells], function(x) {
+    if (is.null(x)) res <- NULL
+    else res <- x$elementId
+
+    if (is.null(res)) res <- paste0("widget", floor(runif(1, max = 1e9)))
```
timelyportfolio: should probably use `htmlwidgets::createWidgetId` [line](https://github.com/ramnathv/htmlwidgets/blob/master/R/widgetid.R#L24) but it unfortunately is not exported

R/combineWidgets.R
```diff
@@ -0,0 +1,186 @@
...
+  # create widget
+  combinedWidget <- htmlwidgets::createWidget(
+    name = 'combineWidgets',
+    x,
+    width = width,
+    height = height,
+    package = 'htmlwidgets'
+  )
+
+  deps <- lapply(widgets, function(x) {
```
timelyportfolio: not huge time savings but reducing by class first and then looking up dependencies should save us some time here.  I'll try to write the code once I finish a first pass.

jbkunst:
Hi,

I make some few tests using the `combineWidgets` in differents widgets and works fine :wink: (not `flexdashboard` yet) but I miss the responsive behavior, @cuche27 are in your plans add this feature?

cuche27:
It looks like it works with flexdashboard too. @jbkunst, what do you mean by "responsive behavior" ? Changing the layout depending on the size of the screen ?

timelyportfolio:
@jjallaire, @ramnathv, @jcheng5 I love this, but I am concerned about adding this to the `htmlwidgets` package.  Here are my questions:

- What happens if someone does agree with this approach or prefers another method?  Do we add another combineWidget function in the package to accommodate?  Accepting this implementation implies that we all agree this is the best and most optimal way to combine widgets now and going forward.
- Other layout helpers for `htmlwidgets` and `htmltools` all exist in their own standalone packages.
- Do we expect to support this function and insure that all `htmlwidgets` correctly work when used with combined widgets?

One advantage I think we get from `combineWidgets` is we can use `saveWidget` on a combination of `htmlwidgets` and can avoid something like my hacky `save_tags` function [lines](https://github.com/timelyportfolio/buildingwidgets/blame/6c86d4b377da4f53fc1739c637a301c336985047/week43/index.Rmd#L22-L38).  Here is one example https://github.com/environmentalinformatics-marburg/mapview/issues/35 where `combineWidgets` would work really well.

I hope my questions do not minimize how helpful this function is.  Great work @cuche27 !!!

jjallaire:
@timelyportfolio You make a strong point about accepting this PR sending a permanent signal that this is "the way" to combine widgets. Maybe it should exist for a time as a standalone package (but still with prominent links from our documentation) before we consider merging it. We could add a page to our docs on "Combining Widgets" and then link to:

1. This package

2. The flexdashboard package

3. The new trelliscopejs package (http://ryanhafen.com/blog/trelliscopejs)

timelyportfolio:
@jbkunst, I think the issue is caused because `width` and `height` are set to `NULL` which means the container for `combineWidgets` gets a fixed size of `960px x 500px`.  Since the size is explicitly provided, a `resize` is not triggered.  Setting `width = "100%"` or `height="100%"` will get closer but is not perfect.  Here is an example.

```
library(combineWidgets)
library(plotly)

combineWidgets(
  title = "The Iris dataset",
  plot_ly(iris, x = ~Sepal.Length, type = "histogram", nbinsx = 20),
  plot_ly(iris, x = ~Sepal.Width, type = "histogram", nbinsx = 20),
  plot_ly(iris, x = ~Petal.Length, type = "histogram", nbinsx = 20),
  plot_ly(iris, x = ~Petal.Width, type = "histogram", nbinsx = 20),
  width = "100%"
)
```

combinewidget_resize

jbkunst:
@timelyportfolio 

Thanks,

If I'm not wrong, this is a `fluid` behavior (I test the `width = 100%` too), but if you force the page wide to a thinner one the `div`s don't stack.

jcheng5:
First of all, this is very cool, @cuche27.

That said, I'd agree that it's too soon to put this in htmlwidgets itself. In particular, I have a feeling that this is a specific scenario of a more general pattern, of producing composite HTML artifacts at the R console/in R scripts and showing them in the Viewer.

```r
library(shiny)
library(htmltools)
library(plotly)

browsable(
  fillPage(
    fillRow(
      plot_ly(iris, x = ~Sepal.Length, type = "histogram", nbinsx = 20),
      plot_ly(iris, x = ~Sepal.Width, type = "histogram", nbinsx = 20)
    )
  )
)
```

(I'm not suggesting that users should be doing this, it's too verbose; rather that combineWidgets might be implemented using this mechanism instead of actually building a true widget just to serve as the layout.)

This style allows the combination of any HTML tags (or objects that can be converted to tags using as.tags), including ones that carry their own HTML dependencies along. What it doesn't give you today is nice "filling" behavior when appropriate (Viewer) and natural fixed-height behavior otherwise (Rmd doc, Shiny app), and it also doesn't force the children to fill their containers. But solving these issues doesn't seem out of line with what we've already done.

ramnathv:
Sorry for jumping on this one late, but I agree that while this is really a cool way to combine widgets @cuche27, I think it belongs in a helper package and `htmlwidgets` should stay agnostic. There are many different ways to achieve widget composition and I agree with @jjallaire that we should prominently feature these approaches and packages, but not become wedded to any one in particular. 

@cuche27 this is really cool work! I have some thoughts and comments on it and will add them sometime next week.

cuche27:
Ok, then it will be a package! :)

I will work a little bit on the documentation and I will submit it to the CRAN quite soon. Then we will see what do the users think.

Do not hesitate to create issues with all your suggestions and comments. And in case you would like to contribute, well it would be really awesome !

Issue #179, opened 2016-01-20
Dependency ordering


cpsievert:
I'd like to use `htmlwidgets::createWidget(dependencies = ...)` to add a conditional dependency and I need that dependency to be placed above "official" dependencies declared in the YAML. Is there a slick way to do this that I'm unaware of?

timelyportfolio:
Not that I know of considering [lines](https://github.com/ramnathv/htmlwidgets/blob/master/R/htmlwidgets.R#L173-L175).  A not slick way is to wrap in a `tagList` and declare the dependencies there.

timelyportfolio:
Even less slick is to leave the `YAML` blank and declare all dependencies with `createWidget(dependencies=...)`.

timelyportfolio:
or get return value from `htmltools::as.tags(plot_ly(...))` and then order the dependencies, so something like

```
p <- plot_ly(...)
p_tag <- htmltools::as.tags(p)

attr(htmltools::as.tags(pc),"html_dependencies") <- rev(attr(htmltools::as.tags(pc),"html_dependencies"))
```

ramnathv:
Just so I understand @cpsievert, you can use the `dependencies` argument in `createWidget`, but the trouble is that it would place the dependency at the end of the other official dependencies. Is that correct?

cpsievert:
yep

ramnathv:
Okay. I don't see any other ways of doing it other than the ones proposed by @timelyportfolio. However, this use case does motivate the need to provide more control for placement of js/css assets. One idea I have toyed around with is introducing a `priority` argument in `htmltools::htmlDependency`. It would function like a `z-index` in html. I am not very clear how priorities would interact across widgets. @jjallaire @jcheng5 any thoughts on this?

timelyportfolio:
Could we just provide a key/priority as @ramnathv suggests, and then sort by the key, so [lines](https://github.com/ramnathv/htmlwidgets/blob/16bd872cfec19c026278ec57b0dc3647b583c6b9/R/htmlwidgets.R#L209-L212
) become?  This would require a change to `htmltools::htmlDependency` also though.

```
dependencies <- c(widget_dependencies(class(x)[1], attr(x, 'package')),
  x$dependencies)

priority <- unique(
  unlist(
    lapply(
      dependencies,
      function(dep, i){
        if(is.null(dep$priority)){
          i
        } else as.numeric(dep$priority)
      },
      # more robust way of doing but for now just
[...]

PR #251, opened 2016-12-09
Move Shiny dependency rendering to be earlier

jcheng5:
The previous ordering caused errors when a widget's JS binding's factory
method (or initialize method, for legacy-style bindings) depends on a
dynamic dependency.

You might wonder why a dependency that's used in the intitialize method
would be dynamic and not static. For my case, it's Crosstalk, which
isn't available as a static dependency; you use crosstalk::crosstalkLibs
from R to attach the deps. It would be nice if there was a way to
express this in the widget YAML, you could just point to an R
package::function() or package::variable (though I suspect some hacks
may be needed to get around R CMD check, which wants all Imports to be
used I think?).

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -489,6 +489,7 @@
       // supported natively by Shiny at the time of this writing.
 
       shinyBinding.renderValue = function(el, data) {
+        Shiny.renderDependencies(data.deps);
         // Resolve strings marked as javascript literals to objects
         if (!(data.evals instanceof Array)) data.evals = [data.evals];
         for (var i = 0; data.evals && i < data.evals.length; i++) {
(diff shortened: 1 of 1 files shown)


jcheng5:
@jjallaire It would be really nice to have this change on CRAN before crosstalk goes to CRAN. It's not absolutely necessary, but it simplifies the widget authors' instructions a bit (they just need to add `dependencies=crosstalk::crosstalkLibs()` to `createWidget()`, instead of both that and hacking up their `mywidgetOutput` to attach the dependencies there too).

```r
# This is necessary either way
myWidget <- function(data, width = NULL, height = NULL) {
  # Some logic goes here...
  
  createWidget("myWidget", x, width=width, height=height,
    dependencies = crosstalk::crosstalkLibs()
  )
}

# With this PR, we don't need this anymore
myWidgetOutput <- function(outputId, width = "100%", height = "400px") {
  htmltools::attachDependencies(
    htmltools::tagList(
      shinyWidgetOutput(outputId, "myWidget", width, height, package = "myWidget")
    ),
    crosstalk::crosstalkLibs()
  )
}
```

jjallaire:
@jcheng5 The last CRAN submission was Nov 9th so we are clear to do another submission. Are you highly confident that this has low regression risk? (i.e. should I just go ahead and submit now)

timelyportfolio:
Should reference https://github.com/ramnathv/htmlwidgets/issues/179.

PR #255, opened 2017-01-18
Add widgetDependencies() S3 generic

cpsievert:
## Summary

- Adds `widgetDependencies()` as a public S3 generic with `character` and `htmlwidget` methods
- The `htmlwidget` method returns **all** dependencies (YAML-declared + JS binding + runtime `dependencies` slot), fixing under-reporting when deps are added dynamically via `createWidget(dependencies = ...)`
- `toHTML()` and `shinyRenderWidget()` now use `widgetDependencies(x)` as the single canonical path for resolving widget dependencies
- `getDependency()` is preserved as a deprecated internal alias

## Motivation

Previously there was no public API to get the complete set of HTML dependencies from a widget object. `getDependency(name, package)` only returned the YAML + binding deps, missing any runtime dependencies added under the `dependencies` slot. This made it difficult for downstream code to introspect widget dependencies accurately.

## Breaking changes

- `getDependency()` now emits a deprecation warning directing users to `widgetDependencies()`
- The first argument is now named `x` (positional callers are unaffected)


ramnathv:
This looks good @cpsievert. Let me test through a couple of scenarios before I merge it in.

PR #258, opened 2017-02-03
Fixes #257

### LINKED ISSUE #257: HTML misinterpreted as code in selfcontained mode
The children of a nested HTML tag gets rendered as code instead of literal HTML when using `prependContent()` and `saveWidget(..., selfcontained = T)`

```r
library(htmltools)
library(htmlwidgets)

one <- tags$div("Some text")
two <- tags$div(one)
p <- prependContent(leaflet::leaflet(), two)
saveWidget(p, "index.html")
```

screen shot 2017-02-02 at 5 45 20 pm

Interestingly enough this same example works as expected when `selfcontained = FALSE`

```r
saveWidget(p, "index.html", selfcontained = FALSE)

# this also works
p <- prependContent(leaflet::leaflet(), one)
saveWidget(p, "index.html")
```

So, perhaps this is a pandoc issue?

cpsievert:
Closes #257

### DIFF
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -32,6 +32,11 @@ pandoc_self_contained_html <- function(input, output) {
   # (note there is no markdown in the source document but
   # we still need to do this "conversion" to get the
   # base64 encoding)
+
+  # in markdown, four spaces means "take this code literally"
+  input_txt <- sub("^\\s+", "", readLines(input))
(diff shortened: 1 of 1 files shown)


cpsievert:
I think https://github.com/ramnathv/htmlwidgets/pull/401 is the better way to approach this

PR #259, opened 2017-02-03
Respect an 'after_widget' field to control the order of dependencies

cpsievert:
See also https://github.com/rstudio/htmltools/pull/76

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -206,13 +206,22 @@ toHTML <- function(x, standalone = FALSE, knitrOptions = NULL) {
       )
     }
   )
-  html <- htmltools::attachDependencies(html,
-    c(widget_dependencies(class(x)[1], attr(x, 'package')),
-      x$dependencies)
-  )
+
+  # make sure we're working with a list of dependencies
+  if (inherits(x$dependencies, "html_dependency")) {
+    x$dependencies <- list(x$dependencies)
+  }
+
+  # collect and attach dependencies
+  deps <- widget_dependencies(class(x)[1], attr(x, 'package'))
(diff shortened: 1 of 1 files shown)


jcheng5:
I was going to suggest adding a separate field into the widget object (in addition to `dependencies`) for dependencies that are prerequisites. But are there cases where you have a prerequisite dependency that isn't know until runtime? (I.e. can't go into the yaml)

cpsievert:
Yeah. IIRC, that was a roadblock for https://github.com/ropensci/plotly/pull/411 since mathjax needed to appear before the plotly.js bundle.

jjallaire:
Another way to solve this is to just forgo YAML dependencies entirely and
define all of them in the order required in the dependencies argument.

cpsievert:
@jjallaire it seems as though forgoing yaml dependencies entirely can lead to weird loading issues (they differ in certain contexts?) -- see https://github.com/ropensci/plotly/issues/1033. Any idea on the best way to proceed?

jjallaire:
I am unsure about how YAML dependencies would differ from explicitly specified ones. @ramnathv or @jcheng5 any idea?

ramnathv:
@jjallaire @jcheng5 I think the issue is that `htmlwidgets` always looks for a YAML file containing dependencies. We [can update it](https://github.com/ramnathv/htmlwidgets/blob/34026bdf095e53ee5cbfaa08379e740dae8efa75/R/utils.R#L48-L50) to check for a YAML file and fetch dependencies only if it contains one. This should solve the problem encountered. I will start a new branch with the fix so @cpsievert can try it out.

PR #265, opened 2017-03-31
Avoid setting .Random.seed to NULL

kbroman:
- Related to Issue [#980](https://github.com/rstudio/rmarkdown/issues/980) for the [rmarkdown](https://github.com/rstudio/rmarkdown) package, `htmlwidgets::createWidgetId()` can cause `.Random.seed` to be set to `NULL`, leading to an odd warning

       '.Random.seed' is not an integer vector but of type 'NULL', so ignored

 - Need to apply to `createWidgetId` the same change that was made to `setWidgetIdSeed` in [commit 7454758](https://github.com/ramnathv/htmlwidgets/commit/7454758) from @kevinushey.

### DIFF
--- a/R/widgetid.R
+++ b/R/widgetid.R
@@ -33,7 +33,10 @@ createWidgetId <- function(bytes = 10) {
     # Change our own seed to match the current seed
     .globals$idSeed <- .GlobalEnv$.Random.seed
     # Restore the system seed--we were never here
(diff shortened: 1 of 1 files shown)


jcheng5:
LGTM. @ramnathv?

cpsievert:
Would love to see this PR merged. These false positive warnings are quite confusing (and rightly so!) for new users.

PR #270, opened 2017-04-25
Don't call `structure(NULL, *)`, as this gives a warning in R 3.4.0.

### LINKED ISSUE #269: NULL values in shinyRenderWidget warn in R 3.4.0
If you use an htmlwidget render function with Shiny on R 3.4.0, you'll get a warning like this if the instance value is `NULL`:

```
Warning in structure(x, class = unique(c("AsIs", oldClass(x)))) :
  Calling 'structure(NULL, *)' is deprecated, as NULL cannot have attributes.
  Consider 'structure(list(), *)' instead.
```

This is originating in the `JSEvals` function, which calls `I(NULL)`. However a lot of the code before it also interacts with the instance value, fortunately it's mostly `$` calls which are safe to call on `NULL`. We might consider doing more-explicit handling of the `NULL` case further upstream and leave `JSEvals` the way it is (or do both).

Repro:

```r
library(shiny)
library(leaflet)

ui <- fluidPage(
  leafletOutput("map")
)

  output$map <- renderLeaflet(NULL)
}

shinyApp(ui, server)
```

bborgesr:
This can be fixed by having the underlying object default to `list()` if it was NULL. Fixes #269

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -141,6 +141,9 @@ JS <- function(...) {
 #   JavaScript are to be identified
 # @author Yihui Xie
 JSEvals <- function(list) {
+  # the `%||% list()` part is necessary as of R 3.4.0 (April 2017) -- if `evals`
+  # is NULL then `I(evals)` results in a warning in R 3.4.0. This is circumvented
+  # if we let `evals` be equal to `list()` in those cases
   evals <- names(which(unlist(shouldEval(list)))) %||% list()
   I(evals)  # need I() to prevent toJSON() from converting it to scalar
 }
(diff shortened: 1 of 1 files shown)


R/utils.R
```diff
@@ -141,7 +141,7 @@ JS <- function(...) {
 #   JavaScript are to be identified
 # @author Yihui Xie
 JSEvals <- function(list) {
-  evals <- names(which(unlist(shouldEval(list))))
+  evals <- names(which(unlist(shouldEval(list)))) %||% list()
```
wch: I think this should have a comment explaining why `list()` is there; otherwise it'll probably be forgotten.

bborgesr:
Closes rstudio/shiny#1682

jcheng5:
@ramnathv @jjallaire This probably deserves a CRAN release, IMHO.

PR #277, opened 2017-06-01
Don't fetch widget dependencies if no yaml file exists

### LINKED ISSUE #179: Dependency ordering

cpsievert:
As @ramnathv points out in https://github.com/ramnathv/htmlwidgets/pull/259#issuecomment-305044861, this makes `getDependency()` only return the htmlwidgets.js dependency if no yaml config file exists.

This should fix #179 and also makes #259 obsolete

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -48,18 +48,16 @@ getDependency <- function(name, package = name) {
     src = system.file("www", package = "htmlwidgets"),
     script = "htmlwidgets.js"
   )
-  
-  config <- sprintf("htmlwidgets/%s.yaml", name)
-  jsfile <- sprintf("htmlwidgets/%s.js", name)
+
+  config <- system.file(sprintf("htmlwidgets/%s.yaml", name), package = package)
+  jsfile <- system.file(sprintf("htmlwidgets/%s.js", name), package = package)
 
   # do less magic if no yaml file exists
-  if (!file.exists(config)) {
+  if (identical(config, "")) {
     return(list(htmlWidgetDep))
   }
-  
-  config <- yaml::yaml.load_file(
-    system.file(config, package = package)
-  )
+
+  config <- yaml::yaml.load_file(config)
   widgetDep <- lapply(config$dependencies, function(l){
     l$src = system.file(l$src, package = package)
     do.call(htmlDependency, l)
@@ -73,7 +71,7 @@ getDependency <- function(name, package = name) {
     if (packageVersion('htmltools') < '0.3.3') {
       bindingDir <- tempfile("widgetbinding")
       dir.create(bindingDir, mode = "0700")
-      file.copy(system.file(jsfile, package = package), bindingDir)
+      file.copy(jsfile, bindingDir)
     } else argsDep <- list(all_files = FALSE)
   }
   bindingDep <- do.call(htmlDependency, c(list(
--- a/R/utils.R
+++ b/R/utils.R
@@ -52,7 +52,10 @@ getDependency <- function(name, package = name) {
   config <- system.file(sprintf("htmlwidgets/%s.yaml", name), package = package)
   jsfile <- system.file(sprintf("htmlwidgets/%s.js", name), package = package)
 
-  # do less magic if no yaml file exists
+  # htmlwidget authors may need to place dependencies *before* the binding
+  # (i.e., jsfile) at print time. So, if no yaml exists, we just include the
+  # htmlwidgets.js dependency and leave it to the author to declare their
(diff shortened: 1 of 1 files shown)


jjallaire:
LGTM. @ramnathv ?

jcheng5:
Looks like we lose bindingDep too in this case--did you intend that?

cpsievert:
> Looks like we lose bindingDep too in this case--did you intend that?

Yes, if we were to include `bindingDep` at this stage, I don't think it would be possible to place other assets before the binding, which doesn't seem right for most cases.

I guess this would introduce breaking changes for widgets that don't have no yaml and no external dependencies (just the binding), but that seems like a fairly small set of packages, and a fix would be easy.

ramnathv:
Apologies for a delayed response on this. @cpsievert on the issue of including the binding dependency, can you throw some light on some use cases. Are there any reasons for wanting to include the binding dependency at a position other than right at the end or at the top?

cpsievert:
I've added a better comment about the early return in  https://github.com/ramnathv/htmlwidgets/pull/277/commits/1c1ace4b8ca006af99960c4cce859fc80d72d990

I need this for a few different things in plotly. For instance, our [TypedArray polyfill](https://github.com/plotly/plotly.js/blob/master/dist/extras/typedarray.min.js) and [MathJaX](https://github.com/plotly/plotly.js/tree/master/dist/extras/mathjax) extras (if included) need to appear before the plotly.js bundle/binding. Furthermore, I'd like to determine whether these (and other) dependencies are really needed at print time, rather than always including them.

ramnathv:
Thanks @cpsievert. This makes sense. Can we also add some documentation around this? You can use the motivating use case for this feature for explaining how to do it.

cpsievert:
Sure, feel free to suggest a place to document.

Before we move forward though, I'm not quite sure this is working in a shiny context. Here is a MRE

```r
devtools::install_github('ropensci/plotly')

library(shiny)
library(plotly)

ui <- fluidPage(
  plotlyOutput("p")
)

server <- function(input, output, ...) {
  output$p <- renderPlotly({
    qplot(1:10)
  })
}

shinyApp(ui, server)
```

Nothing appears, and if you look in `<head>`, none of the plotly dependencies are included, which is odd, especially considering if you set a breakpoint [here](https://github.com/ramnathv/htmlwidgets/blob/387508c5ac162f423fe586533431b267f9883b4d/R/htmlwidgets.R#L436-L437) and call `renderFunc()` the dependencies seem to be there...any ideas what might be happening?

cpsievert:
FWIW, it seems as though [`ShinyBinding.renderValue`](https://github.com/ramnathv/htmlwidgets/blob/387508c5ac162f423fe586533431b267f9883b4d/inst/www/htmlwidgets.js#L491-L518) isn't being called, which is probably the source of the problem

cpsievert:
Interestingly enough, adding the yaml back (as done in https://github.com/ropensci/plotly/commit/45933a55f0ef346d3846106da89207d75f40aaa2) causes `ShinyBinding.renderValue` to be called, which fixes the rendering problem in shiny, but reintroduces the dependency ordering problem.

cpsievert:
Having a closer look at `htmlwidgets.js`, it seems as though there is a general assumption widget binding(s) are known (i.e., registered) before any messages are received from the shiny server.

Perhaps it'd be possible to dynamically register htmlwidget bindings in shiny, but that will likely require significant revisions...

jcheng5:
Why can't the bindingDep be the first thing that's loaded in your case? I.e. don't lose the bindingDep in the no-yaml case. When the binding script executes it should generally just be registering closures, not actually using d3 or whatever...?

jcheng5:
> For instance, our TypedArray polyfill and MathJaX extras (if included) need to appear before the plotly.js bundle/binding. 

This could cause an issue if you have two plotly plots on the same Shiny page that have different dynamic requirements. How bad is it if those extras are loaded after the plotly bundle?

cpsievert:
> why can't the bindingDep be the first thing that's loaded in your case?

Sorry, I should've mentioned that it can (at least it seems to work fine for me at the moment). I'll close the issue (at least for now) since I'm not sure if these proposed changes are a good idea anymore.  

> How bad is it if those extras are loaded after the plotly bundle?

Bad, especially for MathJaX, but perhaps I could export some helpers to insert dependencies in the UI (e.g., `use_mathjax()`/`use_typedarray()`)

PR #287, opened 2017-10-19
Add support for project templates

javierluraschi:
[RStudio project templates](https://blog.rstudio.com/2017/09/13/rstudio-v1.1---the-little-things/#project-templates) provide an easy way for users to create custom R projects. While developing an htmlwidget, I noticed this project could use a custom project template.

Here is how this works, creating a new project will list a new "R Package using HTMLWidgets"

screen shot 2017-10-18 at 5 16 50 pm

In this example, we can create a project named`testwidget`:

screen shot 2017-10-18 at 5 27 39 pm

Then that's it! A project gets created and running the widget produces a simple `1 2 3` custom `<div/>` element. I intentionally kept this widget simple to help users identify their components and customize with ease:

screen shot 2017-10-18 at 5 28 09 pm


jjallaire:
The only thing you might consider here is actually calling the `htmlwidgets::createWidget` function (which would keep you in sync with whatever the latest and greatest boilerplate is). This too generates a simple "hello, world" text widget. It also has a feature to download dependencies from Bower which I guess you could add to the wizard UI as properties.

ramnathv:
@javierluraschi I love the idea and I agree with @jjallaire that it would be good to use `createWidget` and provide a field for widget dependencies (if we can make it searchable from bower, that would be super awesome!).

javierluraschi:
Added a textbox for bower and making use of `scaffoldWidget()`. The textbox can't search for bower packages since, currently, RStudio templates have limited support of UI elements. Here is how the project page looks like with this change:

screen shot 2017-12-14 at 4 45 56 pm

javierluraschi:
@ramnathv anything else that we need in this PR? Thank you!

ramnathv:
@javierluraschi I will take a look tonight and merge it. Thanks!

javierluraschi:
@ramnathv could we merge? Each time that I create an `htmlwidget`, I find myself wanting this feature. Today is one of those days, thank you!

ramnathv:
@javierluraschi I will review tomorrow and merge by Monday! Thanks.

PR #293, opened 2017-12-06
unbreak saveWidget( ... selfcontained = FALSE ...)

ThomasSiegmund:
#292 broke the selfcontained = FALSE case, writing only markdown without conversation to html. This restores the previous behaviour.

### DIFF
--- a/R/savewidget.R
+++ b/R/savewidget.R
@@ -27,13 +27,13 @@ saveWidget <- function(widget, file, selfcontained = TRUE, libdir = NULL,
       sep = "")
   }
 
-  # Save the file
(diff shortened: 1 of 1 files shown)


jjallaire:
@jcheng5 Could you review this?

jcheng5:
Ouch, sorry for introducing that bug! This looks good. Perhaps we should include the <title> in the self-contained=FALSE case as well. I think that should be as simple as `html <- tagList(tags$head(tags$title(title)), html)`?

ThomasSiegmund:
Good point, I have added the title

jcheng5:
Thanks @ThomasSiegmund!

Issue #305, opened 2018-03-26
Using without package?


nstrayer:
Is it possible to build an HTML widget without needing it to be its own package? 

I am currently using HTML widgets to build custom visualization tools for proprietary medical data and have run into the issue of deploying to RStudio Connect as it needs packrat to be able to find the package source. What I'm planning to do is just convert it to a plain shiny module, but I will miss some of the nice scaffolding features that widgets provide. 

I apologize if this has been answered somewhere already, my searches came up empty. 

Thanks for a wonderful package!

timelyportfolio:
We started down this path a while ago in https://github.com/ramnathv/htmlwidgets/issues/176 but never resolved.  Glad you are reviving the discussion.

nstrayer:
I need to familiarize myself with the internals but could this be accomplished by providing a path to libraries and to the main js file to a constructor which returns a list of the main function and the two shiny functions? I'm assuming the package build step is there to gather and inject the JavaScript to a single source?

timelyportfolio:
@nstrayer I'd like to pursue this further as it relates to @fbreitweiser #304 and could provide a good mechanism for quick inclusion of `htmlwidget` functionality.  From what I remember, one of the primary reasons for requiring a full package was the desire for local `dependencies` in a knowable location.  However, with many packages, such as `d3r` and `rmarkdown` offering dependencies and the difficulty specifying order of dependencies in the `yaml` configuration, I am not sure this is as necessary as we originally thought.

Really, if we remove the dependency requirement, we should be able to achieve this objective (unless I am missing something - likely).  So, hopefully this will push the conversation forward.

There is one place where we go looking for file dependencies
- [createWidget lines](https://github.com/ramnathv/htmlwidgets/blob/master/R/htmlwidgets.R#L209-L212) where we call [`widget_dependencies()` lines] which ultimately depends on (https://github.com/ramnathv/htmlwidgets/blob/master/R/htmlwidgets.R#L237-L239) [`getDependency()` lines](https://github.com/ramnathv/htmlwidgets/blob/master/R/utils.R#L41-L80).

One option would be to look for the files as we currently do, but if they do not exist then proceed without error assuming the user or widget builder is intentionally breaking the convention.

jjallaire:
I think you could create htmlwigets outside of a package except for the requirement of the YAML file. If this was optional rather than required then any dependencies could also be provided via the dependencies argument to `createWidget`

timelyportfolio:
[...]

Issue #304, opened 2018-03-22
Don't require YAML file


fbreitwieser:
If the file `htmlwidgets/WIDGET.yaml` does not exist, the following error is thrown:
```
Error in yaml.load(readLines(con), error.label = error.label, ...) : 
  argument "error.label" is missing, with no default
```

However, since dependencies can be specified with `createWidgets`, no yaml file is required.

ramnathv:
@fbreitwieser I would like to understand your use case better. The `dependencies` argument in `createWidgets` was added to support dynamic dependencies. Are you encountering a situation where all dependencies are dynamic and hence you don't want to specify a YAML file?

![image](https://user-images.githubusercontent.com/346288/37792929-68e7ba40-2dca-11e8-8c1e-c719aae0b0b2.png)

fbreitwieser:
Thanks for the quick answer. Personally I just found it easier to define all dependencies using the `dependencies` argument - even the static ones. When using `d3r` for example, I'd define the d3 dependencies with `d3r::d3_dep_v4()`.

ramnathv:
@fbreitwieser That makes sense. It should not be hard modifying the logic so that when a widget yaml file is absent, things work differently.

timelyportfolio:
closed by #306; feel free to reopen for discussion or issues

PR #306, opened 2018-03-29
creation without package or without js binding or yaml config

timelyportfolio:
@nstrayer (#305) and @fbreitwieser (#304) motivated this pull request.  While I think this functionality is for a unique and special case, I very much like the idea of ad hoc widget creation without a package or full set of conventions.

### Base Example
For instance, we can do this which is sort of worthless but provides a foundation for later examples.

```
createWidget("awidget", x=NULL)
```

### Limited Functionality but a Widget Still

```
library(htmltools)
library(htmlwidgets)

browsable(
  tagList(
    tags$script(
"
HTMLWidgets.widget({
  name: 'awidget',
  type: 'output',
  factory: function(el, width, height) {
    return {
      renderValue: function(x) {
        el.innerText = x.message;
      },
      resize: function(width, height) { }
    };
  }
});
"
    ),
    createWidget("awidget", x=list(message="I don't need a package, yaml, or js."))
  )
)
```

### Widget With External Dependency

```
library(htmltools)
library(htmlwidgets)
library(d3r)

browsable(
  tagList(
    tags$script(
"
HTMLWidgets.widget({
  name: 'awidget',
  type: 'output',
  factory: function(el, width, height) {
    return {
      renderValue: function(x) {
        var box = d3.select(el).append('svg').append('rect')
          .attr('height', x.size)
          .attr('width', x.size)
          .style('fill', '#c6b')

        setInterval(
          function() {
            var size = box.attr('height') == x.size ? 100 : x.size
            box.transition()
              .attr('height', size)
              .attr('width', size)
          },
          1000
        )
      },
      resize: function(width, height) { }
    };
  }
});
"
    ),
    createWidget("awidget", x=list(size = 300)),
    d3_dep_v4()
  )
)
```
htmlwidget

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -59,26 +59,23 @@ getDependency <- function(name, package = name){
     })
   }
 
-  bindingDir <- system.file("htmlwidgets", package = package)
-  argsDep <- NULL
-  copyBindingDir <- getOption('htmlwidgets.copybindingdir', TRUE)
-  # TODO: remove this trick when htmltools >= 0.3.3 is on CRAN
-  if (copyBindingDir) {
-    if (packageVersion('htmltools') < '0.3.3') {
-      bindingDir <- tempfile("widgetbinding")
-      dir.create(bindingDir, mode = "0700")
-      file.copy(system.file(jsfile, package = package), bindingDir)
-    } else argsDep <- list(all_files = FALSE)
+  # if js binding does not exist then assume provided through
+  #  some other mechanism such as a specified `htmlDependency` or `script` tag.
+  #  Note, this is a very special case.
+  bindingDep <- NULL
+  if(file.exists(system.file(jsfile, package = package))) {
+    bindingDir <- system.file("htmlwidgets", package = package)
+    argsDep <- NULL
+    bindingDep <- do.call(htmlDependency, c(list(
+      paste0(name, "-binding"), packageVersion(package),
+      bindingDir, script = basename(jsfile)
+    ), argsDep))
   }
-  bindingDep <- do.call(htmlDependency, c(list(
-    paste0(name, "-binding"), packageVersion(package),
-    bindingDir, script = basename(jsfile)
-  ), argsDep))
 
   c(
     list(htmlDependency("htmlwidgets", packageVersion("htmlwidgets"),
-      src = system.file("www", package="htmlwidgets"),
-      script = "htmlwidgets.js"
+                        src = system.file("www", package="htmlwidgets"),
+                        script = "htmlwidgets.js"
     )),
     widgetDep,
     list(bindingDep)
--- a/R/utils.R
+++ b/R/utils.R
@@ -45,13 +45,19 @@ getDependency <- function(name, package = name){
   config = sprintf("htmlwidgets/%s.yaml", name)
   jsfile = sprintf("htmlwidgets/%s.js", name)
 
-  config = yaml::yaml.load_file(
-    system.file(config, package = package)
-  )
-  widgetDep <- lapply(config$dependencies, function(l){
-    l$src = system.file(l$src, package = package)
-    do.call(htmlDependency, l)
-  })
+  # if yaml does not exist then assume no dependencies
+  #  in this cases dependencies should be provided through the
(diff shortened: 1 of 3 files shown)


jjallaire:
LGTM

nstrayer:
When I tried out this it works with your example code but when I attempt to use an arrow function I get this error in the chrome console debugger:
```js
renderValue: function(x) {
      const scatter_data = HTMLWidgets.dataframeToD3(x.data);
      x.domain(d3.extent(scatter_data, d =&gt; d.x));
      y.domain(d3.extent(scatter_data, d =&gt; d.y));
    ....
```

from the R code
```r

devtools::install_github('timelyportfolio/htmlwidgets@freedom', force = TRUE)
library(tidyverse)
library(htmltools)
library(htmlwidgets)
library(d3r)

data <- data_frame(x = rnorm(50), y = rnorm(50))
browsable(
  tagList(
    tags$script(
      "
      HTMLWidgets.widget({
      name: 'awidget',
      type: 'output',
      factory: function(el, width, height) {
      const svg = d3.select(el).append('svg');
      const x = d3.scaleLinear().range([0, width]);
      const y = d3.scaleLinear().range([height, 0]);
      return {
      renderValue: function(x) {
      const scatter_data = HTMLWidgets.dataframeToD3(x.data);
      x.domain(d3.extent(scatter_data, d => d.x));
      y.domain(d3.extent(scatter_data, d => d.y));
      
      svg.selectAll('.scatter_points')
         .data(scatter_data)
         .enter().append('circle')
         .attr('x', d => x(d.x))
         .attr('y', d => y(d.y))
         .attr('r', 5);

      },
      resize: function(width, height) { }
      };
      }
      });
      "
    ),
    createWidget("awidget", x=list(data = data, size = 300)),
    d3_dep_v4()
  )
)
```
So some encoding issues of inline js. Is there an easy way to include external javascript files? Would make this easier to get nice syntax highlighting etc but have everything work without having to spin up a server to serve external js scripts.

timelyportfolio:
@nstrayer wrapping with `HTML()` like `tags$script(HTML(...))` should solve the problem with `=>`, but this might be someday change https://github.com/rstudio/htmltools/issues/75.

timelyportfolio:
@ramnathv @jcheng5 @jjallaire  I can push the button, but I'd prefer some code review or at least an agreement on merge.  I will increment version and update `NEWS`.  Also, where do we stand on next CRAN release?

ramnathv:
This LGTM. I have not had the chance to test it out, but if everyone is comfortable, I am good to merge this.

jjallaire:
Yes, merge away!

jcheng5:
Sorry, I didn't reply but I did review and LGTM as well.

timelyportfolio:
I will pretty up with `NEWS` and minor version bump tonight.  Then I will merge.  Thanks!

R/utils.R
```diff
@@ -45,34 +45,37 @@ getDependency <- function(name, package = name){
...
+    argsDep <- NULL
+    bindingDep <- do.call(htmlDependency, c(list(
+      paste0(name, "-binding"), packageVersion(package),
+      bindingDir, script = basename(jsfile)
+    ), argsDep))
   }
-  bindingDep <- do.call(htmlDependency, c(list(
-    paste0(name, "-binding"), packageVersion(package),
-    bindingDir, script = basename(jsfile)
-  ), argsDep))
```
yihui: The rewrite here is not equivalent to the original code. It threw away `argsDep <- list(all_files = FALSE)`, and the consequence is that everything under the `htmlwidgets/` directory (e.g., https://github.com/rstudio/DT/tree/master/inst/htmlwidgets) will be copyid to `*-binding/` when the widget is rendered, which is not desirable -- only `jsfile` should be copied. Originally this was controlled by the parameter `all_files = FALSE`.

yihui: BTW, I noticed this issue when I was trying to rebuild the DT website with the latest version of htmlwidgets, and a lot of new files showed up in GIT.

yihui:
Sure. I'll provide a fix in a few minutes. Actually this is not too bad. It only copies several unused files, and won't break anything.

PR #312, opened 2018-04-26
Restore the all_files = FALSE argument in htmlDependency()

yihui:
When creating the HTML dependency for the JS binding script, we need to make sure only this single script is copied when the widget is rendered (all_files = TRUE is the default, meaning all files/dirs will be copied, which is undesirable).

The issue was introduced in #306.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -62,14 +62,12 @@ getDependency <- function(name, package = name){
   # if js binding does not exist then assume provided through
   #  some other mechanism such as a specified `htmlDependency` or `script` tag.
   #  Note, this is a very special case.
-  bindingDep <- NULL
-  if(file.exists(system.file(jsfile, package = package))) {
+  bindingDep <- if (file.exists(system.file(jsfile, package = package))) {
     bindingDir <- system.file("htmlwidgets", package = package)
-    argsDep <- NULL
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -62,14 +62,12 @@ getDependency <- function(name, package = name){
...
+  bindingDep <- if (file.exists(system.file(jsfile, package = package))) {
     bindingDir <- system.file("htmlwidgets", package = package)
-    argsDep <- NULL
-    bindingDep <- do.call(htmlDependency, c(list(
+    htmlDependency(
       paste0(name, "-binding"), packageVersion(package),
-      bindingDir, script = basename(jsfile)
-    ), argsDep))
+      bindingDir, script = basename(jsfile), all_files = FALSE
+    )
```
yihui: To reviewers: `x <- NULL; if (FALSE) { x <- ...}` is equivalent to `x <- if (FALSE) { ... }`, because `if (FALSE) {...}` returns `NULL`, and `if (TRUE) { foo }` returns `foo`.

timelyportfolio:
Thanks @yihui, I am sorry I missed that.  I should have known better.  I am glad you discovered quickly and really appreciate the quick fix.

yihui:
@timelyportfolio No worries!

PR #315, opened 2018-05-17
bindings.resize no longer called unnecessarily on first window resize

jankurianski:
To reproduce:
1. Open http://www.htmlwidgets.org/showcase_dygraphs.html
2. Open DevTools
3. Set a breakpoint in `htmlwidgets.js` line `582`: `binding.resize(el, size.w, size.h, initResult);`
4. Refresh the page
5. Resize the window
6. Observe your breakpoint gets hit, even though the widget didn't change size

This fix records the initial widget size on the first render so that the first resize event doesn't fire the resize binding.  The motivation to fix this is not for the sake of optimization, but rather to improve any widget which has resize logic that performs large visual changes like animations, etc.

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -565,7 +565,10 @@
         }
 
         if (binding.resize) {
-          var lastSize = {};
+          var lastSize = {
+            w: sizeObj ? sizeObj.getWidth() : el.offsetWidth,
+            h: sizeObj ? sizeObj.getHeight() : el.offsetHeight
+          };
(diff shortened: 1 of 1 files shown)


jjallaire:
@jcheng5 You have the most context on how resize works and whether this might introduce any unanticipated side effects. Could you review?

jcheng5:
(Sorry for the delay, I missed the notification.)

jcheng5:
@jjallaire Do you think it's worth adding a NEWS item?

jjallaire:
Yeah, let's add a NEWS item.

PR #324, opened 2019-01-09
Allow yaml dependencies to be function calls

jcheng5:
This PR makes it possible for the dependencies .yaml file to point to `htmlDependency` objects that come from other packages, instead of just ones that are bundled in this package.

```
dependencies:
  - call: shiny::bootstrapLib()
```

- [ ] Example
- [ ] Unit tests
- [ ] NEWS


timelyportfolio:
@jcheng5 I agree this would be a nice addition.  Please let me know how I might help.

PR #327, opened 2019-02-04
Try evaling tasks with and without parentheses, closes #326

### LINKED ISSUE #326: onRender should support more than a bare function
I'd expect this to work, but it currently doesn't:

```r
library(plotly)
library(reactR)

plot_ly() %>%
    onRender(babel_transform("gd => console.log(gd)"))
```

That's because `babel_transform()` doesn't return a bare function:

```r
cat(babel_transform("gd => console.log(gd)"))
```

```js
"use strict";

(function (gd) {
  return console.log(gd);
});
```

However, `HTMLWidgets.evalAndRun` assumes we're passing a bare function:

https://github.com/ramnathv/htmlwidgets/blob/5db27ada3895b8cf2ba6ee1dd9ebfa5201729e5b/inst/www/htmlwidgets.js#L236

Seems as though we could support this use case by also trying `var taskFunc = eval(task);`

cpsievert:
See #326

### DIFF
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -233,7 +233,17 @@
           theseArgs = theseArgs.concat([task.data]);
           task = task.code;
         }
-        var taskFunc = eval("(" + task + ")");
+        try {
+          var taskFunc = eval("(" + task + ")");
+        } catch(err) {
+          try {
+            var taskFunc = eval(task);
+          } catch(error) {
+            console.error(err);
+            console.error(error);
+          }
+        }
+
         if (typeof(taskFunc) !== "function") {
           throw new Error("Task must be a function! Source:\n" + task);
         }


inst/www/htmlwidgets.js
```diff
@@ -233,7 +233,17 @@
...
         }
-        var taskFunc = eval("(" + task + ")");
+        try {
+          var taskFunc = eval("(" + task + ")");
+        } catch(err) {
+          try {
+            var taskFunc = eval(task);
+          } catch(error) {
+            console.error(err);
+            console.error(error);
```
jcheng5: I don't think we should silently swallow here

jcheng5:
This isn't a good idea--if the eval'd code has side effects, they could happen more than once. If you really need this, I think it'd be better to introduce either an explicit argument switching the behavior, or a new function altogether.

cpsievert:
Closing in favor of #328

PR #328, opened 2019-02-04
add validate argument to onRender() to optionally ensure server-side that jsCode can evaluate to a function, fixes #326

### LINKED ISSUE #326: onRender should support more than a bare function

cpsievert:


R/htmlwidgets.R
```diff
@@ -152,22 +152,42 @@ appendContent <- function(x, ...) {
...
+      codeType2 <- try(ctx$eval("typeof eval('(' + code + ')')"), silent = TRUE)
+      if (identical(codeType2, "function")) {
+        jsCode <- paste0("(", jsCode, ")")
+      } else {
+        stop("The `jsCode` argument does not evaluate to a function.")
+      }
+    }
+  } else {
+    jsCode <- paste0("(", jsCode, ")")
+  }
```
cpsievert: @alandipert makes a good point why this is a bad idea -- `jsCode` might contain an expression that only works in a browser environment

PR #332, opened 2019-03-12
Try evaling js code client-side both with and without parentheses

### LINKED ISSUE #329: onRender() doesn't support expressions with a trailing comma
From the documentation for `onRender()`:

> The jsCode parameter must be a valid JavaScript expression that returns a function.

This works as expected

```r
library(plotly)
library(htmlwidgets)

plot_ly() %>%
  onRender("(function (gd) { return console.log(gd); })")
```

But, if you have a (valid) expression with a trailing semicolon, it doesn't work:

```r
library(plotly)
library(htmlwidgets)

plot_ly() %>%
  onRender("(function (gd) { return console.log(gd); });")
```

The main use case for supporting the trailing comma is to make it easier to take the output of something like `reactR::babel_transform()` and feed it into `onRender()`

cpsievert:
Fixes #329 

TODO: 
 - [x] Consider this eval strategy for `JS()` as well?

https://github.com/ramnathv/htmlwidgets/blob/5db27ada3895b8cf2ba6ee1dd9ebfa5201729e5b/inst/www/htmlwidgets.js#L735


inst/www/htmlwidgets.js
```diff
@@ -233,7 +233,21 @@
           theseArgs = theseArgs.concat([task.data]);
           task = task.code;
         }
-        var taskFunc = eval("(" + task + ")");
+        var taskFunc = null;
+        try {
+          taskFunc = eval(task);
+        } catch(error) {
+          if (!error instanceof SyntaxError) {
+            throw new Error(error.message);
```
jcheng5: Try `throw error;` instead, so we preserve the class and other properties (and stack trace too, I believe?).

inst/www/htmlwidgets.js
```diff
@@ -233,7 +233,21 @@
...
+          }
+          // To support function declarations, retry evaluation with
+          // parentheses (turns declarations into an expression)
+          try {
+            taskFunc = eval("(" + task + ")");
+          } catch(e) {
+            throw new Error(error.message);
```
jcheng5: If and only if `e instanceof SyntaxError` do we `throw error;`. Otherwise, we `throw e;`.

jcheng5:
Looks good!

The only downside I can think of is if you have a real syntax error in your old-style JS (i.e. the one that does require us to add parens) you'll only ever see the syntax error that arises from not having parens, not the true error message. But I don't know how we can do better without parsing, and hopefully people can figure out other ways of diagnosing their syntax errors.

jcheng5:
Do you mind adding a test case to shiny-examples that has both types of callbacks? (Prefer if you didn't use Babel, just hardcode the output, so we don't all have to install V8 for testing)

cpsievert:
Great, test added in https://github.com/rstudio/shiny-examples/pull/129

BTW, @jcheng5, you'll have to merge as I don't write access

PR #345, opened 2019-09-20
Schedule staticRender to run after initShiny

cpsievert:
See https://github.com/rstudio/shiny/issues/2630

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: htmlwidgets
 Type: Package
 Title: HTML Widgets for R
-Version: 1.4
+Version: 1.5
 Authors@R: c(
     person("Ramnath", "Vaidyanathan", role = c("aut", "cph")),
     person("Yihui", "Xie", role = c("aut")),
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -660,7 +660,11 @@
   }
 
   // Wait until after the document has loaded to render the widgets.
-  if (document.addEventListener) {
+  if (shinyMode && window.jQuery.fn.jquery > "3.0.0") {
+    window.jQuery(function() {
+      setTimeout(window.HTMLWidgets.staticRender, 1)
+    });
+  } else if (document.addEventListener) {
     document.addEventListener("DOMContentLoaded", function() {
       document.removeEventListener("DOMContentLoaded", arguments.callee, false);
       window.HTMLWidgets.staticRender();
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -660,7 +660,12 @@
   }
 
   // Wait until after the document has loaded to render the widgets.
-  if (shinyMode && window.jQuery.fn.jquery > "3.0.0") {
+  if (shinyMode && window.jQuery) {
+    // Note: this case is here to ensure that initShiny executes
+    // before staticRender executes. Since initShiny currently
+    // registers important methods like Shiny.onInputChange, if
+    // it happens to execute after staticRender, the widget
+    // won't have access to all of Shiny's methods.
     window.jQuery(function() {
       setTimeout(window.HTMLWidgets.staticRender, 1)
     });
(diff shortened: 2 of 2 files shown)


inst/www/htmlwidgets.js
```diff
@@ -660,7 +660,11 @@
   }
 
   // Wait until after the document has loaded to render the widgets.
-  if (document.addEventListener) {
+  if (shinyMode && window.jQuery.fn.jquery > "3.0.0") {
```
jcheng5: How about:

```
if (shinyMode) {
```

or

```
if (window.jQuery) {
```

? I don't think we ever want to behave differently than jQuery, if jQuery is available.

inst/www/htmlwidgets.js
```diff
@@ -660,7 +660,11 @@
...
+    window.jQuery(function() {
+      setTimeout(window.HTMLWidgets.staticRender, 1)
```
wch: Should have a semicolon here

inst/www/htmlwidgets.js
cpsievert: Wouldn't that change the order of execution if we are outside of Shiny but using jQuery (due to the `setTimeout`)? Maybe it's safer to do:

```
if (shinyMode && window.jQuery) {
```

cpsievert: Now that I think of it, we could also do:

```
if (shinyMode) {
window.jQuery(document).on("shiny:bound", function() {
      window.HTMLWidgets.staticRender();
 });
}
```

Since that's what we'll likely suggest to shiny users?

cpsievert: After talking with @wch and @schloerke we decided it's probably best to not use any of the existing shiny events and instead use `$(fn)`

inst/www/htmlwidgets.js
```diff
@@ -660,7 +660,16 @@
...
   // Wait until after the document has loaded to render the widgets.
-  if (document.addEventListener) {
+  if (shinyMode && window.jQuery) {
+    // Note: this case is here to ensure that initShiny executes
+    // before staticRender executes. Since initShiny currently
+    // registers important methods like Shiny.onInputChange, if
+    // it happens to execute after staticRender, the widget
+    // won't have access to all of Shiny's methods.
+    window.jQuery(function() {
+      setTimeout(window.HTMLWidgets.staticRender, 1)
```
wch: semicolon. :)

inst/www/htmlwidgets.js
```diff
@@ -660,7 +660,16 @@
...
   // Wait until after the document has loaded to render the widgets.
-  if (document.addEventListener) {
+  if (shinyMode && window.jQuery) {
+    // Note: this case is here to ensure that initShiny executes
+    // before staticRender executes. Since initShiny currently
+    // registers important methods like Shiny.onInputChange, if
+    // it happens to execute after staticRender, the widget
+    // won't have access to all of Shiny's methods.
+    window.jQuery(function() {
+      setTimeout(window.HTMLWidgets.staticRender, 1);
```
jcheng5: Are you sure this needs a `setTimeout`? I actually don't think it does. I know that leaflet `updateBounds` function needs `init_shiny` to have completed, but for that reason, leaflet itself calls `setTimeout(updateBounds)`

cpsievert: Yes, plotly needs it; without it, `Shiny.setInputValue()` won't be available when `HTMLWidgets.renderValue()` is called.

jcheng5: @cpsievert And that issue for plotly doesn't repro with the CRAN version of Shiny?

jcheng5: Sorry, I'm not trying to be difficult, just trying to make the most conservative change we can for now.

cpsievert: Oh, it does repro with CRAN versions...I wouldn't be surprised if there are lots of htmlwidgets out there are unknowingly suffering from the same issue.

What's the potential danger in deferring `staticRender()`? Is it that shiny might not be able to bind to the widget?

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: htmlwidgets
 Type: Package
 Title: HTML Widgets for R
-Version: 1.4
+Version: 1.5
```
jcheng5:
```suggestion
Version: 1.4.0.9000
```
I'll change it to 1.5 when I create an RC branch

PR #353, opened 2019-10-07
v1.5.1 release candidate

jcheng5:


inst/www/htmlwidgets.js
```diff
@@ -668,44 +668,24 @@
...
   } else if (document.attachEvent) {
     document.attachEvent("onreadystatechange", function() {
       if (document.readyState === "complete") {
         document.detachEvent("onreadystatechange", arguments.callee);
-        window.HTMLWidgets.staticRender();
+        if (shinyMode && window.jQuery && has_jQuery3()) {
+          window.jQuery(window.HTMLWidgets.staticRender);
+        } else {
+          window.HTMLWidgets.staticRender();
+        }
```
cpsievert: Maybe it's time to drop the `onreadystatechange` shim? https://caniuse.com/#feat=domcontentloaded

jcheng5: Yes but let's do that after we ship 1.5.1?

PR #357, opened 2019-11-07
Try evaluating code with parentheses before without parentheses

### LINKED ISSUE #356: Troubles with the use of dygraphs
With htmlwidgets 1.5.1, there are some big troubles with some functions of the dygraphs package (e.g. dyStackedBarGroup or dyShadow).
Its is OK with the version 1.3.
I don't know if you should fix htmlwidgets or @pshevtsov should adapt dygraphs to version 1.5.1.

```
library(dygraphs)
library(magrittr)

## dataset
lungDeaths <- cbind(fdeaths, mdeaths, ldeaths, foo = fdeaths/2, bar = fdeaths/3)

## A bunch of different plotters together:
dygraph(lungDeaths) %>%
  dyRangeSelector() %>%
  dyBarSeries('bar') %>% 
  dyStemSeries('mdeaths') %>% 
  dyShadow('foo') %>% 
  dyFilledLine('fdeaths')

## Stacked Bar and Ribbon Graphs:  
dygraph(lungDeaths) %>% 
  dySeries('mdeaths', axis = 'y2') %>%
  dyAxis('y', valueRange = c(-100, 1000)) %>% 
  dyStackedBarGroup(c('ldeaths', 'fdeaths'))
```

cpsievert:
Closes #356. 

The problem with evaluating without parentheses first is that named function declarations are valid expressions that return undefined (whereas evaluating with parentheses returns the function, as expected).

It's true there may be weird corner-cases where switching this order leads to something else that's undesirable, but we've been operating with the 'with parentheses' model for so long that switching the ordering now will minimize backward-compatibility issues

### DIFF
--- a/inst/NEWS
+++ b/inst/NEWS
@@ -1,3 +1,8 @@
+htmlwidgets 1.5.1.9000
+-------------------------------------------------------
+
+* Fixed an issue with passing named function declarations to `JS()` and `onRender()` (introduced by v1.4). (#356)
+
 htmlwidgets 1.5.1
 -------------------------------------------------------
 
--- a/inst/www/htmlwidgets.js
+++ b/inst/www/htmlwidgets.js
@@ -249,13 +249,13 @@
   function tryEval(code) {
     var result = null;
     try {
-      result = eval(code);
+      result = eval("(" + code + ")");
     } catch(error) {
       if (!error instanceof SyntaxError) {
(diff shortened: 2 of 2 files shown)


cpsievert:
A test for this functionality is here https://github.com/rstudio/shiny-examples/pull/177

PR #368, opened 2020-05-15
Avoid error in find_pandoc() by replacing file.exists() with dir.exis…

maximilian-leitheiser:
…ts()

Since 'pandoc' is appended to src in find_version(src) before querying for the version, the paths in sources are required to point to the directory of the pandoc executable. Thus, a check by dir.exists(src) instead of the current file.exists(src) seems more appropriate. In particular, it avoids an error if src points the the pandoc executable itself instead of the directory.


cpsievert:
This has been addressed by #443

Issue #376, opened 2020-07-24
*_html protection?


thebioengineer:
Working on developing an html widget and I accidentally started implementing the "*_html" method for creating custom "housings" for widgets.

I was very surprised and thought it was a bug until I found [your documentation](http://www.htmlwidgets.org/develop_advanced.html#custom-html-widget). 

Would it maybe be good to add protections around the calling of the function and ensuring that it returns a valid shiny.tag class and throw a warning? 

For backwards compatibility it might be a nightmare, but could there be a case for converting widget_html into a method/switch where it looks for widget_html.widgetname? that seems like it would be more in-line with standard R practice for different behavior for your custom classes.

```r
widget_html <- function (name, package, id, style, class, inline = FALSE, ...) {
    fn <- tryCatch(get(paste0("widget_html.",name), asNamespace(package), 
        inherits = FALSE), error = function(e) widget_html.default)
    fn(id = id, style = style, class = class, ...)
}

widget_html.default <- function (name, package, id, style, class, inline = FALSE, ...) {
   if (inline) {
        tags$span(id = id, style = style, class = class)
    } else {
        tags$div(id = id, style = style, class = class)
    }
}
```

jcheng5:
Thanks, Ellis--both sound like good suggestions to me. Any chance you're up for preparing a PR?

thebioengineer:
Sure thing @jcheng5 - is there one you prefer? (both could be implemented as well)

jcheng5:
I think they're complementary? The warning to catch if you've stumbled into this accidentally like you did. The S3 method to make it possible to eventually deprecate the old way.

thebioengineer:
Perfect. I'll aim to open a PR tomorrow. Thanks Joe!

cpsievert:
I think this has been addressed by https://github.com/ramnathv/htmlwidgets/pull/378?

PR #377, opened 2020-07-24
Update widget_html

thebioengineer:
Addresses #376

Suggested two updates to widget_html that are implemented here:
 - To behave like an s3 method where it will look up the "type" of method it should use. This uses the "name" field to determine. Minimal changes from current *name*_html method.
 - Add an object output type checker for custom functions to make sure they return a `shiny.tag`. Throws a warning describing what it saw and what it expected. 

We could add a soft-deprication for the current method (*name*_html) if you want. I wasn't sure how we might want to do that. 
- Add a check for the 'current' method and throw a warning? That might alarm users as opposed to the authors though
- Open to ideas!

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -239,15 +239,6 @@ widget_html_wrapper <- function(name, package, id, style, class, inline = FALSE,
     inline = inline,
     ...)
 
-  if(!inherits(fn_res,c("shiny.tag","shiny.tag.list"))){
-    warning(
-      paste0("widget_html.",name),
-      " returned an object of class `",
-      deparse(class(fn_res)),
-      "` instead of a `shiny.tag`."
-    )
-  }
-
   fn_res
 
 }
@@ -265,6 +256,7 @@ widget_html.default <- function (name, package, id, style, class, inline = FALSE
                  error = function(e) NULL)
 
   if (is.function(fn)) {
+
     .Deprecated(
       new = paste0("widget_html.",name),
       package = package,
@@ -274,7 +266,17 @@ widget_html.default <- function (name, package, id, style, class, inline = FALSE
                   "See docs for details at http://www.htmlwidgets.org/develop_advanced.html"),
       old = paste0(name, "_html")
     )
+
     fn(id = id, style = style, class = class, ...)
+
+    if(!inherits(fn_res,c("shiny.tag","shiny.tag.list"))){
+      warning(
+        paste0("widget_html.",name),
+        " returned an object of class `",
+        deparse(class(fn_res)),
+        "` instead of a `shiny.tag`."
+      )
+    }
 
   } else if (inline) {
     tags$span(id = id, style = style, class = class)
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -200,7 +200,7 @@ toHTML <- function(x, standalone = FALSE, knitrOptions = NULL) {
     container(
       htmltools::tagList(
         x$prepend,
-        widget_html(
+        widget_html_wrapper(
           name = class(x)[1],
           package = attr(x, "package"),
           id = id,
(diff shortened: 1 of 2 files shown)


thebioengineer:
Should I add myself as a contributor?

thebioengineer:
@jcheng5 - does this PR satisfy what you were thinking WRT #376? Or should I add in the soft-deprecation of the current behavior?

jcheng5:
Sorry I missed this earlier, @thebioengineer.

I misread your original suggestion in #376, what I thought you were proposing was:

1. Turn `widget_html` into an actual S3 generic
2. Today's `widget_html` becomes `widget_html.default` (with the addition of checking the return value)
3. Any widgets that want to override `widget_html` can either use the old way, or preferably, define a `widget_html.yourclassname` method

This way, there's no breakage in backwards compatibility, and it's using the actual S3 generic dispatch mechanism instead of our own lookalike.

The harder question is when and how we would deprecate the old way, I think a PR like the above would be the first step which would allow us to update our docs. The next step would be to search CRAN reverse imports/dependencies and @timelyportfolio's repos to see if anyone is actually using the old way.

Sorry for the miscommunication!

thebioengineer:
Gotcha! Yeah, we can do that. I'll work on that tonight.

timelyportfolio:
@jcheng5 @thebioengineer I agree this should be `S3` and your idea is a more robust solution.  I do use `*_html` but is an easy change.

thebioengineer:
hey @jcheng5, I update html_widgets to behave like an s3 method. To facilitate the actual s3 method dispatch, I created a wrapper function - `html_widget_wrapper`. It also does the type checking we discussed. 

Not sure I like this solution over our pseudo s3 dispatch method because I now we will have to export the methods, and so will the widget authors. 

I added .Deprecated() to warn users about the changing methodology, but we can remove that if we want to hold off.

I checked this with both {reactable} and {leaflet} and it appeared to work still.

R/htmlwidgets.R
```diff
@@ -228,18 +228,54 @@ toHTML <- function(x, standalone = FALSE, knitrOptions = NULL) {
...
+      new = paste0("widget_html.",name),
+      package = package,
+      msg = paste("Using `widgetname_html` is about to be deprecated.",
+                  "widget_html going forward will now follow standard s3 method",
+                  "deployment for creating custom HTML 'housings'.",
+                  "See docs for details at http://www.htmlwidgets.org/develop_advanced.html"),
+      old = paste0(name, "_html")
+    )
     fn(id = id, style = style, class = class, ...)
+
```
jcheng5, line 278: Can you please move the "x returned an object of class y instead of shiny.tag" check/warning to here? I'd rather not check the results if they've implemented the S3 method, as there are conceivably other things you could return as housing besides tags (`HTML()`, objects coercible using `htmltools::as.tags`).

jcheng5:
Other than my nit picky comment, this looks great now, thanks! I don't see exporting the generic (or `S3method`-ing the methods) as a problem, this is a totally mainstream use of S3, AFAICT.

R/htmlwidgets.R
```diff
@@ -228,18 +228,56 @@ toHTML <- function(x, standalone = FALSE, knitrOptions = NULL) {
 
 }
 
+widget_html_wrapper <- function(name, package, id, style, class, inline = FALSE, ...) {
 
-widget_html <- function(name, package, id, style, class, inline = FALSE, ...){
+  fn_res <- widget_html(
+    name = structure(name, class = c("character", name)),
```
jcheng5, line 234: Sorry, I just noticed this little detail, give me a second to get my head around this.

thebioengineer: AFAIK s3 methods are deployed based on the class of the object into UseMethod. This sets that. Otherwise we need to do our own s3 deployment method.  unless there is another way to decide which s3 method to use?

R/htmlwidgets.R
thebioengineer: sure thing

R/htmlwidgets.R
jcheng5: The problem with setting the classname on the name is that it's a lie; we shouldn't tag objects with class names unless they actually are objects of that class. I didn't look closely enough at this initially and just assumed an `x` object was being passed in and that would be what we use for S3 dispatch. I had forgotten that this `widget_html` can be called before any `x` object exists--namely, in `shinyWidgetOutput`, where we're just generating the markup which won't actually be populated with the widget until the server renders it.

So that being the case, I don't think it makes sense to use actual S3 dispatch, I'm fine with the pseudo-S3 approach.

Sorry for the unnecessary extra work. The first commit was pretty close, I'll cherry-pick from this PR and probably file a new one to hopefully be merged next week.

thebioengineer: No worries, It was a good thought exercise as I went through this. I wasn't super clear about my intention for a "pseudo" s3 method implementation, just that I thought something similar might be more "R-like" and less likely to introduce someone stumbling across the *widgetname*_html method and getting weird results. Thanks for the review Joe!

thebioengineer:
@jcheng5 - assuming this PR can be closed since your branch has been merged into master?

PR #378, opened 2020-08-07
Pseudo S3 widget_html dispatch

jcheng5:
Addresses #376, succeeds #377 

- If `PACKAGE:::widget_html.WIDGETNAME` exists, use it. Else,
- If `PACKAGE:::WIDGETNAME_html` exists, use it, and warn if the results aren't a tag, tag list, or `HTML()`. Else,
- Use `htmlwidgets:::widget_html.default`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -9,6 +9,7 @@ Authors@R: c(
     person("Joe", "Cheng", role = c("aut"), email = "joe@rstudio.com"),
     person("Carson", "Sievert", role = c("aut", "cre"), email = "carson@rstudio.com", comment = c(ORCID = "0000-0002-4958-2844")),
     person("Kenton", "Russell", role = c("aut", "cph")),
+    person("Ellis", "Hughes", role = c("ctb")),
     person(family = "RStudio", role = "cph")
     )
 Description: A framework for creating HTML widgets that render in various
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,7 +24,7 @@ Imports:
     yaml
 Suggests:
     knitr (>= 1.8),
-    rmarkdown
+    rmarkdown,
     testthat
 Enhances: shiny (>= 1.1)
 URL: https://github.com/ramnathv/htmlwidgets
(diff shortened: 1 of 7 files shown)


thebioengineer:
Thanks Joe! Happy to hear my idea was helpful!

jcheng5:
Planning to merge this right after the next CRAN release

jcheng5:
@cpsievert I rebased and force-pushed. If this passes checks it should be good to merge AFAICT.

DESCRIPTION
```diff
@@ -24,6 +25,7 @@ Imports:
 Suggests:
     knitr (>= 1.8),
     rmarkdown
+    testthat
```
cpsievert, line 28:
```suggestion
    rmarkdown,
    testthat
```

vignettes/develop_advanced.Rmd
```diff
@@ -177,12 +177,14 @@ If multiple arguments are passed to `JS()` (as in the above example), they will
...
-sparkline_html <- function(id, style, class, ...){
+widget_html.sparkline <- function(id, style, class, ...){
   tags$span(id = id, class = class)
 }
 ```
 
-Note that this function is looked up within the package implementing the widget by the convention `widgetname_html` so it need not be formally exported from your package or otherwise registered with **htmlwidgets**.
+Note that this function is looked up within the package implementing the widget by the convention <code>widget_html.<em>widgetname</em></code> so it need not be formally exported from your package or otherwise registered with **htmlwidgets**.
+
+(**htmlwidgets** 1.5.1 and earlier used a convention of <code><em>widgetname</em>_html</code>. This is still supported for now, but the new <code>widget_html.<em>widgetname</em></code> convention is recommended going forward, as it seems less likely to lead to false positives.)
```
cpsievert, line 187:
```suggestion
(**htmlwidgets** 1.5.2 and earlier used a convention of <code><em>widgetname</em>_html</code>. This is still supported for now, but the new <code>widget_html.<em>widgetname</em></code> convention is recommended going forward, as it seems less likely to lead to false positives.)
```

cpsievert (approved):
LGTM assuming the tests pass!

PR #382, opened 2020-08-28
RS-6898: Changing htmlwidgets to allow IFRAME-less widgets in Displayr.

oliverbock:


oliverbock:
Whoops, went to wrong repository.

PR #385, opened 2020-09-27
ensure resolveSizing to use only first values from knitrOptions$out.width|height.px (fixes yihui/knitr#1901)

atusy:
This PR solves an error that arises when multiple values are given to `fig.width` or `fig.height`:

> Error in validateCssUnit(sizeInfo$width): CSS units must be a single-element numeric or character vector 

The issue become a obvious problem when using HTML widgets in **flexdashboard** with knitr 1.30.
However, it has been the potential problem in any output formats with older knitr as well.
See https://github.com/yihui/knitr/issues/1901 for details.

@jcheng5 Would you consider CRAN release after merging this PR? This issue affects many users of **flexdashboard**.
Thank you!

## minimal reprex

````
---
output: html_document
---

```{r, fig.width=1:2}
library(dygraphs)

lungDeaths <- cbind(mdeaths, fdeaths)
dygraph(lungDeaths)
```
````

cc. @yihui

### CODE REVIEW

**Comment by jcheng5** [comment]
I didn’t know fig.width and fig.height could have multiple values, what does it mean?

**Comment by atusy** [comment]
Multiple values will be applied to multiple plots, respectively.

````
```{r, fig.width=5:6}
g <- ggplot2::qplot(1,1)

print(g)
print(g)
```
````

In **flexdashboard**, it is also used with combination with multiple values for `fig.ext`.
This is to provide larger images for desktop and smaller images for mobile.

cderv:
@jcheng5 I believe all flexdashboard containing htmlwidgets are currently breaking with the last CRAN version of knitr due a fix that happens in knitr regarding options (See https://github.com/yihui/knitr/issues/1901#issuecomment-701290767 for details)

I think this is something we should fix as soon as possible, including a CRAN release of htmlwidgets. 

If you think this should be dealt elsewhere than this suggested PR, happy to discuss! 

Thanks!

jcheng5:
cc @cpsievert

cpsievert:
Thanks @atusy for looking into this and providing the fix! I've cherry-picked your commit into a release candidate (#387)  that we'll be shipping within the next 24 hours (assuming nothing else pops up in our testing suite)

PR #394, opened 2020-11-19
Update saveWidget(selfcontained=T) to use same relative path fix as h…

### LINKED ISSUE #299: saveWidget to a relative path
Just had the problem to save my widget into a subdirectory relative to my working directory. 
Is there a special reason, why relative paths do not work? I had a question on stack overflow for this https://stackoverflow.com/questions/48690837/saving-interactive-plotly-graph-to-a-path-using-htmlwidget/48691210

So in the end, absolute path works and as I figured out with an older issue here, by using `normalizePath()` it also works. Why not relative path form?

```
results_dir <- 'results'
if(!dir.exists(results_dir)) dir.create(results_dir)
p <- plotly::plot_ly(economics, x = ~date, y = ~pop, 
             type = 'scatter', mode = 'markers')
htmlwidgets::saveWidget(p, 
                        file.path(results_dir, 'VSGs.html'))
```

cpsievert:
For more context, see https://github.com/rstudio/htmltools/pull/192. Closes #299

### DIFF
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -19,9 +19,12 @@ pandoc_save_markdown <- function(html, file, background = "white", title, libdir
 
   # ensure that the paths to dependencies are relative to the base
   # directory where the webpage is being built.
-  dir <- dirname(file)
-  oldwd <- setwd(dir)
-  on.exit(setwd(oldwd), add = TRUE)
+  if (is.character(file)) {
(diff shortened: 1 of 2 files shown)


cpsievert:
Merging since these changes were approved in https://github.com/rstudio/htmltools/pull/192

PR #395, opened 2020-11-21
Remove indentation when rendering HTML into pandoc

### LINKED ISSUE #358: htmlwidgets::saveWidget selfContained adds <pre> tags around custom HTML
The `timevis` package uses the function `timevis_html` to add a custom zoom menu:
https://github.com/daattali/timevis/blob/be9756629bf12f5225d578d10c093d8f968f8551/R/timevis.R#L505

This works fine in combination with htmlwidgets::saveWidget selfContained=FALSE.
When using the selfContained=TRUE option however, the custom HTML is wrapped in `<pre>` tags, and is therefore not present in the output.

As a workaround, the generated HTML code can be adapted afterwards, but this is less than ideal.

Minimal reprex: https://gist.github.com/rfaelens/9dc31900c8e6601f2b978be4f5c4ec43
Referencing original issue: https://github.com/daattali/timevis/issues/87

andrewdenner:
When the html object contains nested tags, the default behaviour is for renderTags to indent the code. This causes an issue when the generated text is parsed by pandoc later on. Pandoc parses the indented block as "Verbatim (code) blocks", escaping it as text in the HTML and wrapping it in `<pre><code>` tags, The result is a non-functional widget.

To avoid this, set indent = FALSE in the call to renderTags, avoiding this opportunity for pandoc to misinterpret the input file. 
Believe that this fixes #358

### DIFF
--- a/R/pandoc.R
+++ b/R/pandoc.R
@@ -26,7 +26,7 @@ pandoc_save_markdown <- function(html, file, background = "white", title, libdir
     on.exit(setwd(owd), add = TRUE)
   }
 
-  rendered <- renderTags(html)
+  rendered <- renderTags(html, indent = FALSE)
 
   deps <- lapply(rendered$dependencies, function(dep) {
     dep <- htmltools::copyDependencyToDir(dep, libdir, FALSE)


cpsievert:
Thanks for the contribution, I think we'll be taking this as a part of #401

PR #401, opened 2021-01-25
Protect HTML from markdown parsing

### LINKED ISSUE #358: htmlwidgets::saveWidget selfContained adds <pre> tags around custom HTML

jcheng5:
This builds on #395, preserving indentation when possible. 

Closes #358
Closes https://github.com/ramnathv/htmlwidgets/issues/257


ismirsehregal:
Why is this PR not merged yet?

PR #435, opened 2022-08-29
adapted to pandoc version >= 2.19 and fixed bug in docs

neuwirthe:
`pandoc 2.19` deprecated command line option `--self-contained` .
Therefore the command line `options` in the call to `pandoc_convert` in `pandoc_self_contained_html` need to be changed to `--embed-resources` and `--standalone`.


cpsievert:
Thanks! I addressed this via https://github.com/ramnathv/htmlwidgets/commit/f7ab7979c95b2dd99edf965ef368cfa449fcb3b8

PR #437, opened 2022-09-27
Add vfill-item CSS class when height isn't specified by user

cpsievert:
With this PR, the htmlwidget's HTML container gains a `.vfill-item` CSS class when `height` is `NULL` (on the widget instance when statically rendered or `shinyWidgetOutput()` when rendered in shiny). 

The primary motivation for this is so widgets can fill as needed inside `bslib::card_body_fill()`  ([#452](https://github.com/rstudio/bslib/pull/452)); but more generally, it allows widgets (with `height` of `NULL`) to fill any HTML element with a `.vfill-container` class (when used with `{bslib}`, since those CSS classes are provided by `bs_theme()`), for example:

```r
library(bslib)
library(htmltools)

browsable(page_fluid(
    theme = bs_theme(),
    style = "height:100vh", 
    class = "vfill-container",
    plotly::plot_ly()
))
```

This means, at least for widgets that follow the [recommended R binding](https://www.htmlwidgets.org/develop_intro.html#r-binding) approach, statically rendered widgets (such as above) won't need any changes to fill by default (within a `vfill-container`). However, the recommend Shiny binding approach of 

```r
sigmaOutput <- function(outputId, width = "100%", height = "400px") {
  shinyWidgetOutput(outputId, "sigma", width, height, package = "sigma")
}
```

will change to 

```r
sigmaOutput <- function(outputId, width = NULL, height = NULL) {
  shinyWidgetOutput(outputId, "sigma", width, height, package = "sigma")
}
```

and default sizing should instead be provided through a CSS file, providing something like:

```css
.sigma.html-widget {
  width: 100%;
  height: 400px;
}
```

## TODO 

Update website (and the sigma package?) to reflect the new default sizing approach

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -27,12 +27,12 @@ jobs:
           - {os: macOS-latest,   r: 'release'}
           - {os: windows-latest, r: 'release'}
           - {os: windows-latest, r: '3.6'}
-          - {os: ubuntu-16.04,   r: 'devel', rspm: "https://packagemanager.rstudio.com/cran/__linux__/xenial/latest", http-user-agent: "R/4.0.0 (ubuntu-16.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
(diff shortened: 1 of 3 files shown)


cpsievert:
Closing in favor of https://github.com/rstudio/shiny/pull/3715

PR #444, opened 2022-10-28
Have widget container default to overflow:hidden when fill=TRUE

cpsievert:
Requires https://github.com/rstudio/htmltools/pull/347

### DIFF
--- a/R/htmlwidgets.R
+++ b/R/htmlwidgets.R
@@ -181,7 +181,12 @@ toHTML <- function(x, standalone = FALSE, knitrOptions = NULL) {
     name, package, id = x$id,
     style = css(
       width = validateCssUnit(sizeInfo$width),
-      height = validateCssUnit(sizeInfo$height)
+      height = validateCssUnit(sizeInfo$height),
+      # bindFillRole() puts `overflow:auto` to items by default,
+      # which is a sensible generic default, but in the context of
+      # widgets, we can pretty easily run into non-pixel-perfect situations
+      # (e.g., the widget JS uses something like offsetHeight to resize itself)
+      "--html-fill-item-overflow" = if (sizeInfo$fill) "hidden"
     ),
     class = paste(name, "html-widget"),
     width = sizeInfo$width,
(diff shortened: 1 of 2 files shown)


inst/www/htmlwidgets.js
```diff
@@ -576,29 +578,30 @@
...
         if (binding.initialize) {
-          initResult = binding.initialize(el,
-            sizeObj ? sizeObj.getWidth() : el.offsetWidth,
-            sizeObj ? sizeObj.getHeight() : el.offsetHeight
-          );
+          initResult = binding.initialize(el, lastSize.w, lastSize.h);
           elementData(el, "init_result", initResult);
         }
 
         if (binding.resize) {
```
jcheng5, line 602: Need to set `lastSize` here, as the size could've changed as a result of initialization

R/htmlwidgets.R
```diff
@@ -182,13 +182,15 @@ toHTML <- function(x, standalone = FALSE, knitrOptions = NULL) {
...
       height = validateCssUnit(sizeInfo$height),
+    ),
+    class = paste(
+      name, "html-widget",
       # bindFillRole() puts `overflow:auto` to items by default,
       # which is a sensible generic default, but in the context of
       # widgets, we can pretty easily run into non-pixel-perfect situations
       # (e.g., the widget JS uses something like offsetHeight to resize itself)
-      "--html-fill-item-overflow" = if (sizeInfo$fill) "hidden"
+      if (sizeInfo$fill) "html-fill-item-overflow-hidden"
```
jcheng5, line 192: I think it might be better to not use `paste()` with conditionals inside; it results in extra spaces:

```r
> paste("1", NULL, "2")
[1] "1  2"
```

It's not that big of a deal but it's also very easy to use the approach where you have multiple `class` arguments.

cpsievert: > it's also very easy to use the approach where you have multiple class arguments.

I don't think we can actually since this function can be provided by widget authors and have both `class` and `...` in the formal arguments (see the bottom of http://www.htmlwidgets.org/develop_advanced.html)

Issue #452, opened 2022-12-11
Pandoc 2.19 has deprecated --self-contained


dmurdoch:
`saveWidget()` calls `pandoc_self_contained_html()` which [calls Pandoc with the `--self-contained` option](https://github.com/ramnathv/htmlwidgets/blob/e375b340140e80ce73db8d0a9dfdb5d3492bc896/R/pandoc.R#L114).  However, Pandoc 2.19 [has deprecated that option](https://pandoc.org/releases.html) in favour of ` --embed-resources --standalone`, and so it gives a warning.  For example, see https://stackoverflow.com/questions/74379298/argument-selfcontained-deprecated-in-htmlwidgetssavewidget .

cpsievert:
This has been addressed via #257

PR #453, opened 2022-12-11
`--self-contained` is deprecated in Pandoc 2.19.

dmurdoch:
This fixes issue #452 .


cpsievert:
This has already been addressed via #443

dmurdoch:
I think that fix got overwritten in 4dd30ca765c20edc8 , so the problem is still there in the master branch.

PR #472, opened 2023-09-07
Remove use of html-fill-item-overflow-hidden class

cpsievert:
This is in response to https://github.com/rstudio/htmltools/pull/401, which removed the `overflow:auto` on fill items (and also removed the `html-fill-item-overflow-hidden` class since it's no longer relevant).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,15 +21,13 @@ URL: https://github.com/ramnathv/htmlwidgets
 BugReports: https://github.com/ramnathv/htmlwidgets/issues
 Imports:
     grDevices,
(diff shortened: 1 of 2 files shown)


DESCRIPTION
```diff
@@ -21,13 +21,15 @@ URL: https://github.com/ramnathv/htmlwidgets
 BugReports: https://github.com/ramnathv/htmlwidgets/issues
 Imports:
     grDevices,
-    htmltools (>= 0.5.4),
+    htmltools (>= 0.5.4.9000),
```
cpsievert, line 24:
```suggestion
    htmltools (>= 0.5.7),
```

PR #473, opened 2023-10-16
POSIXlt should not be evaluated in JS

yihui:
This is to fix rstudio/DT#1092. `POSIXlt` is essentially a list, but we should stop the recursion here.


cpsievert:
Thanks @yihui, we'll take this on in #467

PR #491, opened 2024-12-12
Remove old, problematic, and unnecessary `renderError` and `clearError` method defaults

### LINKED ISSUE #490: Widget HTML containers have `visibility:hidden;` when an error is rendered
Discovered via https://github.com/rstudio/shiny/issues/4115#issuecomment-2538908422

For example, note how, because of the `req()`, no spinner appears. The underlying reason is that htmlwidgets.js adds `visibility:hidden;` to the widget container when the error gets rendered, and doesn't clear that style until the next value gets rendered.

```r
library(shiny)
library(plotly)

ui <- fluidPage(
  useBusyIndicators(),
  actionButton("run_plotly", "Run plotly"),
  plotlyOutput("p")
)

  output$p <- renderPlotly({
    validate(need(input$run_plotly, "Please run plotly"))
    Sys.sleep(3)
    plotly::plot_ly()
  })
}

shinyApp(ui, server)
```

cpsievert:
Closes #490

By removing these defaults, when the widget is in rendered in Shiny, it will simply fallback to Shiny's OutputBinding default methods, which does a better and simpler job of rendering and clearing errors.

And, when rendering statically, the renderError/clearError methods aren't relevant

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # htmlwidgets (development version)
 
-
+* Closes #490: htmlwidgets no longer inserts a new HTML element (and hides the widget HTML element) to display error messages. Instead, it now inserts error messages in the widget container. This is more consistent with how Shiny output bindings work in general, and as a result, allows the new `shiny::useBusyIndicators()` to work as intended when errors (silent or otherwise) are relevant. (#491)  
 
 # htmlwidgets 1.6.4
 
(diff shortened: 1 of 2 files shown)


cpsievert:
After looking at this again, I now realize there is an intentional difference here and the generic output binding `renderError`. In the widget case, when an error gets rendered _after_ a widget has been rendered, the widget container's height doesn't change, meaning there is no reflow of content below it. This isn't the case for something like `uiOutput()`.

We could still move away from the `setTimeout()` to `ResizeObserver()`, but that's another matter.

PR #493, opened 2025-04-20
Add litedown support

dmurdoch:
`litedown` is Yihui's lightweight version of `rmarkdown` + `knitr`.  Whereas there was support for `htmlwidgets` in `knitr`, there's none in `litedown`.  This PR adds it.

It does require a couple of small changes to `litedown` which have been submitted:  https://github.com/yihui/litedown/pull/86 .


jcheng5:
Thanks @dmurdoch! I’m fine with this change in spirit but would prefer to wait until the litedown change is merged.

dmurdoch:
That makes sense.  I just realized one thing:  if there's an htmlwidget that uses an htmlDependency sourced online, it won't work.  I can see how it should be fixed, and will modify soon.

dmurdoch:
I've just added support for `href` dependencies.   I am not sure I understand all the possible forms of `src` in a dependency, and the code doesn't handle them all, but  I was only able to find one package that used `href`, and none of the more exotic possibilities.

PR #498, opened 2026-07-07
Add support for binary compression of widget data

andrjohns:
When creating files with multiple widgets and large input data, the resulting HTML filesize is dominated by the embedded JSON data.

This PR introduces the option (disabled by default): `htmlwidget.compress_payload`, to store widget data as a compressed binary string. Decompression is handled by the [`fflate`](https://github.com/101arrowz/fflate) library - only the [`unzlibSync`](https://github.com/101arrowz/fflate/blob/master/docs/functions/unzlibSync.md) and [`strFromU8`](https://github.com/101arrowz/fflate/blob/master/docs/functions/strFromU8.md) functions are bundled, resulting in only 5kb (minified) added to the output.

In our local testing this reduced the output size for one of our HTML files from 55mb -> 10mb, with little lag in the initial rendering time.

Let me know if I've missed anything, or if there's a better way to do this, thanks!


andrjohns:
Closing as we've implemented via a different route

